PR #19, opened 2018-05-11
Reduce permissions to the max

### LINKED ISSUE #16: Limit the RBAC to only what the operator needs?
I noticed that the rbac rule for this operator allows unlimited access to all the resources as shown here:

```yam
...
rules:
- apiGroups:
  - "*"
  resources:
  - "*"
  verbs:
  - "*"
```
https://github.com/Dynatrace/dynatrace-oneagent-operator/blob/master/deploy/rbac.yaml

It is probably a best practice to limit the access of the operator to the only resources and actions it needs.

An example of that is from the recently open source vault-operator:
https://github.com/coreos/vault-operator/blob/master/example/rbac-template.yaml#L5-L43
you can see that the rbac rule only gives the resource access to what vault-operator needs.

baichinger:
Reduce RBAC permissions to the bare minimum in order to restrict access to
objects operated on by a defined set of operations.

Resolves #16

### DIFF
--- a/deploy/rbac.yaml
+++ b/deploy/rbac.yaml
@@ -8,11 +8,35 @@ metadata:
     operator: oneagent
 rules:
 - apiGroups:
-  - "*"
+  - dynatrace.com
   resources:
-  - "*"
+  - oneagents
   verbs:
-  - "*"
+  - get
+  - list
+  - watch
+  - update
+- apiGroups:
+  - apps
+  - extensions
+  resources:
+  - daemonsets
+  verbs:
+  - get
+  - list
+  - watch
+  - create
+  - update
+  - delete
+- apiGroups:
+  - ""
+  resources:
+  - pods
+  verbs:
+  - get
+  - list
+  - watch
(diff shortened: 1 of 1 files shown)


deploy/rbac.yaml
```diff
@@ -8,11 +8,35 @@ metadata:
...
+  - daemonsets
+  verbs:
+  - get
+  - list
+  - watch
+  - create
+  - update
+  - delete
+- apiGroups:
+  - ""
```
fanminshi: Maybe add `# "" indicates the core API group` here for readability?

deploy/rbac.yaml
```diff
@@ -8,11 +8,35 @@ metadata:
...
+  - oneagents
   verbs:
-  - "*"
+  - get
+  - list
+  - watch
+  - update
+- apiGroups:
+  - apps
+  - extensions
```
fanminshi: is there a reason to use `extensions`? here?

PR #29, opened 2018-06-28
Secure access token via secrets

baichinger:
Plain text access tokens in CR got replaced by a secret holding `apiToken` and
`paasToken`. Name of the secret to use is depicted by `.spec.tokens`. If this
field is empty the name of the CR is used to look up the secret.

### DIFF
--- a/pkg/stub/handler.go
+++ b/pkg/stub/handler.go
@@ -55,8 +55,26 @@ func (h *Handler) Handle(ctx types.Context, event types.Event) error {
 			return err
 		}
 
+		// get tokens for api access
+		if oneagent.Status.Tokens == "" || oneagent.Spec.Tokens != "" && oneagent.Status.Tokens != oneagent.Spec.Tokens {
+			if oneagent.Spec.Tokens != "" {
+				oneagent.Status.Tokens = oneagent.Spec.Tokens
+			} else {
+				oneagent.Status.Tokens = oneagent.Name
+			}
+			updateStatus = true
+		}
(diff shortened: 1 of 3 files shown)


pkg/stub/handler.go
```diff
@@ -55,8 +55,26 @@ func (h *Handler) Handle(ctx types.Context, event types.Event) error {
 			return err
 		}
 
+		// get tokens for api access
+		if oneagent.Status.Tokens == "" || oneagent.Spec.Tokens != "" && oneagent.Status.Tokens != oneagent.Spec.Tokens {
```
MartinGoodwell: can we have a variable for each of the conditions?
eg *oneagent.Spec.Tokens != ""* is checked again in the next if. would improve readability a lot

pkg/stub/handler.go
```diff
@@ -210,6 +230,38 @@ func updateDaemonSet(oa *v1alpha1.OneAgent) error {
...
+	err := query.Get(obj)
+	if err != nil {
+		logrus.WithFields(logrus.Fields{"oneagent": cr.Name, "error": err, "secret": cr.Name}).Error("failed to get secret")
+		return "", err
+	}
+
+	value, ok := obj.Data[key]
+	if !ok {
+		logrus.WithFields(logrus.Fields{"oneagent": cr.Name, "secret": cr.Name, "key": key}).Warning("key not found")
+		return "", nil
```
esukram: This case is not ideal, as there is a potential failure hidden.

PR #31, opened 2018-07-05
Synchronize changes between CR and DS

baichinger:
Add support for updating a DaemonSet in case a change was made either in the
OneAgent custom resource or in the DaemonSet itself.

* Refactor logic for building DaemonSet objects.

* Move default value computation to defaulter funcs.

* Refactor computation of ONEAGENT_INSTALLER_TOKEN and purge dependency on
  .status.tokens.

### DIFF
--- a/pkg/apis/dynatrace/v1alpha1/defaults.go
+++ b/pkg/apis/dynatrace/v1alpha1/defaults.go
@@ -1,5 +1,12 @@
 package v1alpha1
 
+import (
+	"fmt"
+	"strconv"
+
+	corev1 "k8s.io/api/core/v1"
+)
+
 func SetDefaults_OneAgentSpec(obj *OneAgentSpec) {
 	if obj.WaitReadySeconds == nil {
 		obj.WaitReadySeconds = new(uint16)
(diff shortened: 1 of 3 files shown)


pkg/stub/handler.go
```diff
@@ -203,22 +210,35 @@ func getPodReadyState(p *corev1.Pod) bool {
...
-// updateDaemonSet creates a new DaemonSet object if it does not exist.
+// upsertDaemonSet creates a new DaemonSet object if it does not exist or
+// updates an existing one if changes need to be synchronized.
 //
 // Returns an error in the following conditions:
 //  - all k8s apierrors except IsNotFound
 //  - failure on daemonset creation
-func updateDaemonSet(oa *v1alpha1.OneAgent) error {
+func upsertDaemonSet(oa *v1alpha1.OneAgent) error {
 	ds := getDaemonSet(oa)
```
MartinGoodwell: ds --> daemonSet?
maybe, for better readability?

pkg/stub/handler.go
```diff
@@ -203,22 +210,35 @@ func getPodReadyState(p *corev1.Pod) bool {
...
+			}
+		}
 		return nil
-	}
-
-	if apierrors.IsNotFound(err) {
+	} else if apierrors.IsNotFound(err) {
+		// create deamonset
 		logrus.WithFields(logrus.Fields{"oneagent": oa.Name}).Info("deploying daemonset")
+		desired := oa.DeepCopy()
```
MartinGoodwell: desiredWhat?

pkg/stub/handler.go
esukram: The very short variable names are pretty common in go-lang.

I would leave it with `ds`.

pkg/apis/dynatrace/v1alpha1/defaults.go
```diff
@@ -1,5 +1,12 @@
 package v1alpha1
 
+import (
+	"fmt"
+	"strconv"
+
+	corev1 "k8s.io/api/core/v1"
+)
+
 func SetDefaults_OneAgentSpec(obj *OneAgentSpec) {
```
esukram: The `obj` should be `cr` or something similar.

pkg/apis/dynatrace/v1alpha1/defaults.go
```diff
@@ -13,4 +20,28 @@ func SetDefaults_OneAgentSpec(obj *OneAgentSpec) {
 	if len(obj.Args) == 0 {
 		obj.Args = append(obj.Args, "APP_LOG_CONTENT_ACCESS=1")
 	}
+
+	if _, ok := obj.NodeSelector["beta.kubernetes.io/os"]; !ok {
```
esukram: Is this sufficient? What if the user specifies multiple node selectors, does this still work as desired? So, is `NodeSelector` as map?

pkg/apis/dynatrace/v1alpha1/defaults.go
baichinger: Right, but the stub for this function got generated by `defaulter-gen`. The name `obj` is used throughout the whole Kubernetes source:
https://github.com/search?q=org%3Akubernetes+%22func+SetDefaults_%22&type=Code

pkg/apis/dynatrace/v1alpha1/defaults.go
baichinger: `NodeSelector` is of type `map[string]string`. Other selectors do not get overwritten.

pkg/stub/handler.go
```diff
@@ -252,6 +248,18 @@ func upsertDaemonSet(oa *v1alpha1.OneAgent) error {
 	return nil
 }
 
+// hasSpecChanged compares essential OneAgent custom resource settings with the
+// actual settings in the DaemonSet object
+func hasSpecChanged(dsSpec *appsv1.DaemonSetSpec, cr *v1alpha1.OneAgent) bool {
```
esukram: This guy has high potential for _unit testing_ :)

Further more putting into a util class could be also considered...

PR #38, opened 2018-07-27
Add unit tests and create operator specific package

baichinger:
- Refactor handler and create package for oneagent handler functionality.
- Provide unit tests.


pkg/apis/dynatrace/v1alpha1/oneagent.go
```diff
@@ -0,0 +1,46 @@
...
+	}
+}
+
+// Validate sanity checks if essential fields in the custom resource are available
+//
+// Return an error in the following conditions
+// - ApiUrl empty
+func (oa *OneAgent) Validate() error {
+	var msg []string
+	if len(oa.Spec.ApiUrl) == 0 {
```
esukram: Based on https://stackoverflow.com/a/18595217/9897333 - I would also recommend using `oa.Spec.ApiUrl == ""`.

pkg/oneagent/oneagent.go
```diff
@@ -0,0 +1,237 @@
...
+	"k8s.io/apimachinery/pkg/labels"
+)
+
+// time between consecutive queries for a new pod to get ready
+const splayTimeSeconds = uint16(10)
+
+// Reconcile reconciles the OneAgent DaemonSets to the spec specified by
+// oneagent custom resource. State changes are detected by comparing spec
+// fields from custom resource and its related DaemonSet.
+func Reconcile(oneagent *v1alpha1.OneAgent) error {
```
esukram: Just a note: this function is rather long, maybe worth considering a split...

pkg/oneagent/oneagent.go
```diff
@@ -0,0 +1,237 @@
...
+			return err
+		}
+
+		// wait for pod on node to get "Running" again
+		var status error
+		fieldSelector, _ := fields.ParseSelector(fmt.Sprintf("spec.nodeName=%v,status.phase=Running,metadata.name!=%v", pod.Spec.NodeName, pod.Name))
+		labelSelector := labels.SelectorFromSet(cr.BuildLabels())
+		logrus.WithFields(logrus.Fields{"field-selector": fieldSelector, "label-selector": labelSelector}).Debug("query pod")
+		listOps := &metav1.ListOptions{FieldSelector: fieldSelector.String(), LabelSelector: labelSelector.String()}
+		for splay := uint16(0); splay < *cr.Spec.WaitReadySeconds; splay += splayTimeSeconds {
```
esukram: `for` within `for` makes it hard follow, rethink if we can refactor this handling to a function to get it less complex.

pkg/runtime/v1alpha1/oneagent.go
```diff
@@ -0,0 +1,207 @@
...
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// Validate sanity checks if essential fields in the custom resource are available
+//
+// Return an error in the following conditions
+// - ApiUrl empty
+func Validate(oa *api.OneAgent) error {
```
esukram: Wouldn't that method make more sense to a `method` alike `os.Validate()`?

pkg/util/util.go
```diff
@@ -0,0 +1,64 @@
...
+		"oneagent":  name,
+	}
+}
+
+// getPodReadyState determines the overall ready state of a Pod.
+// Returns true if all containers in the Pod are ready.
+func GetPodReadyState(p *corev1.Pod) bool {
+	ready := true
+	for _, c := range p.Status.ContainerStatuses {
+		ready = ready && c.Ready
```
esukram: You can short cut the loop, if you receive a `false`.

pkg/runtime/oneagent.go
```diff
@@ -0,0 +1,244 @@
...
+		if err := waitPodReadyState(pod, cr); err != nil {
+			logrus.WithFields(logrus.Fields{"oneagent": cr.Name, "nodeName": pod.Spec.NodeName, "warning": err}).Warning("timeout waiting on pod to get ready")
+			return err
+		}
+	}
+
+	return nil
+}
+
+func waitPodReadyState(pod corev1.Pod, cr *api.OneAgent) error {
```
esukram: The signature is on opposite order as compared to the above one.

PR #42, opened 2018-08-20
Added Dockerfile-rhel which uses rhel-atomic as base image

jsm84:


MartinGoodwell:
Hi @jsm84,
sorry, I missed one thing: the Dockerfile references a license file, which is not part of the PR.
Would you mind providing that?
Thanks,
Martin

jsm84:
The "LICENSE" file is located at the root of the repository, so it's not necessary.Of course, not having it in the same directory as "Dockerfile-rhel" means that you couldn't call docker build from that directory. You couldn't anyway in this case, since the sole build artifact (operator binary) is missing or exists outside of the build context dir of ./tmp/build.In order to build the container successfully, you would call "docker build" like so:`docker build -t <tag> -f tmp/build/Dockerfile-rhel ./` from the root of the source repo. I can document this in a README.md file and submit that (since this is how it would have to be built if not using the operator-sdk utility).On Aug 22, 2018 10:47, Martin Goodwell <notifications@github.com> wrote:Hi @jsm84,
sorry, I missed one thing: the Dockerfile references a license file, which is not part of the PR.
Would you mind providing that?
Thanks,
Martin

Issue #56, opened 2018-10-25
Changing the group of the custom resource


heydenb:
We want to use this operator, but our customer is only allowing definitions being deployed by the CD system.
This user has only rights to deploy resources if they are in group "thecompany.com"

So, in stead of
`kind: CustomResourceDefinition
metadata:
  name: oneagents.dynatrace.com
spec:
  group: dynatrace.com
...` 
We do
`kind: CustomResourceDefinition
metadata:
  name: oneagents.thecompany.com
spec:
  group: thecompany.com
...` 

Deploying the custom resource works, but the operator looks for "oneagents.dynatrace.com" and you can't configure that. Or so it seems.
Would it be possible to make that configurable via a parameter passed to the operator

aloismayr:
Hi @heydenb 
I understand the request and that it would be an easy approach for making the operator available in the CD controlled environment. However, we would not like to make the identity of the operator configurable as outlined in the description. Mainly for two reasons:

1) This not only has an impact to the operator itself but also all the deployed custom resources. It will also touch the lifecycle (upgrades, etc.) of the operator. 
2) It seems the main purpose of this feature request is to "by-pass" the security policy of the CD environment, as this specific user is only allowed to deploy resources of group "thecompany.com". Why can't they just allow to also run resources of group "dynatrace.com"?

esukram:
Closing this issue alongside the PR #57 as undesired.

PR #57, opened 2018-11-08
Make group configurable

heydenb:
I believe these changes are enough to solve 
https://github.com/Dynatrace/dynatrace-oneagent-operator/issues/56

Please review and merge into the next release so this change is supported and we can use the official image


esukram:
Undesired behavior.

PR #66, opened 2019-01-25
Introduce option to disable automatic agent update restarts

baichinger:
OneAgent pods get restarted in case a new version gets published on the cluster. Option `disableAgentUpdate` disables this behavior if set to `true`.


pkg/controller/oneagent/oneagent_controller.go
```diff
@@ -143,17 +143,22 @@ func (r *ReconcileOneAgent) Reconcile(request reconcile.Request) (reconcile.Resu
...
-		err := r.updateCR(instance)
+	if instance.Spec.DisableAgentUpdate == true {
+		reqLogger.Info("automatic oneagent update is disabled")
+		return reconcile.Result{}, nil
+	} else {
+		updateCR, err = r.reconcileVersion(reqLogger, instance)
 		if err != nil {
 			return reconcile.Result{}, err
-		}
+		} else if updateCR == true {
```
delMar43: is `== true` really necessary?

PR #75, opened 2019-03-12
Update minimum requirements for the Operator

lrgar:
The Operator SDK v0.4 requires Kubernetes 1.11+ (hence OCP 3.11+ as well.)

See: https://github.com/operator-framework/operator-sdk/blob/v0.4.x/doc/user-guide.md#prerequisites

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,8 +21,10 @@ From here on Dynatrace OneAgent Operator controls the lifecycle and keeps track 
 ## Supported platforms
 
 Dynatrace OneAgent Operator is supported on the following platforms:
-* Kubernetes 1.9+
-* OpenShift Container Platform 3.9+
+* Kubernetes 1.11+
+* OpenShift Container Platform 3.11+
+
+For Kubernetes 1.9+ and OpenShift 3.9+, the Dynatrace Operator v0.2.0 can be used.
 
 Help topic _How do I deploy Dynatrace OneAgent as a Docker container?_ lists compatible image and OneAgent versions in its [requirements section](https://www.dynatrace.com/support/help/infrastructure/containers/how-do-i-deploy-dynatrace-oneagent-as-docker-container/#requirements).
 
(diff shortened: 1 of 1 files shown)


esukram:
We should provide something like a table, with the supported versions.

There is high chance, that certain breaking changes will be introduced, making it easier for our users to know which Operator version for which Kubernetes / Openshift version.

baichinger:
But why? Works great on OCP 3.9.

lrgar:
On OCP 3.10, if you delete the Operator pod, the leadership lock (a configmap) doesn't get removed automatically. Once the pod is recreated, the Operator will wait until such lock gets deleted to do anything, and the user would need to do it manually to make it work.

On my testing, the automatic deletion worked on Kubernetes 1.11+/OCP 3.11.

Since the Operator SDK 0.4 dependency already requires Kubernetes 1.11+-based environments, it's probably better to just limit ourselves to that.

README.md
```diff
@@ -20,9 +20,12 @@ From here on Dynatrace OneAgent Operator controls the lifecycle and keeps track
...
 ## Supported platforms
 
-Dynatrace OneAgent Operator is supported on the following platforms:
-* Kubernetes 1.9+
-* OpenShift Container Platform 3.9+
+Depending of the version of the Dynatrace OneAgent Operator, it supports the following platforms:
+
+| Dynatrace OneAgent Operator version | Kubernetes | OpenShift Container Platform |
+| ----------------------------------- | ---------- | ---------------------------- |
+| master                              | 1.11+      | 3.11+                        |
```
esukram: v0.3.0+ / master
or something?

lrgar: I'll add an v0.3.0 row once we release it properly.

baichinger:
A build on _master_ currently creates an image with tag _snapshot_. Suggest to change the image tag to _master_ as  it would streamline terms in documentation and manifests.

lrgar:
> A build on _master_ currently creates an image with tag _snapshot_. Suggest to change the image tag to _master_ as it would streamline terms in documentation and manifests.

Makes sense to me. Will handle it.

PR #77, opened 2019-03-12
Fixes whitespace issue with token, causing them to be invalid

namratachaudhary:


pkg/controller/oneagent/util.go
```diff
@@ -127,7 +127,7 @@ func getToken(secret *corev1.Secret, key string) (string, error) {
 		return "", err
 	}
 
-	return string(value), nil
+	return strings.TrimSpace(value), nil
```
esukram: One could use https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/api/core/v1/types.go#L4992
`Secret.stringData` to have the `string` representation.

baichinger: `StringData` is write-only and only used for creating secrets. Don't see viable alternatives to `strings.TrimSpace` or `bytes.TrimSpace`.

esukram: Wasn't aware this fact.

pkg/controller/oneagent/util.go
```diff
@@ -127,7 +127,7 @@ func getToken(secret *corev1.Secret, key string) (string, error) {
 		return "", err
 	}
 
-	return strings.TrimSpace(value), nil
+	return strings.TrimSpace(string(value)), nil
```
esukram: Works.

pkg/controller/oneagent/util.go
namratachaudhary: My IDE failed to give this error on compilation. Fixed it, should be able to catch it next time.

namratachaudhary:
@esukram @lrgar Added tests, there's a comment [here](https://github.com/Dynatrace/dynatrace-oneagent-operator/pull/77/files#diff-d95cdfb368554bd6faec472e6089b802R72) where I had doubt, please check.

esukram:
> @esukram @lrgar Added tests, there's a comment [here](https://github.com/Dynatrace/dynatrace-oneagent-operator/pull/77/files#diff-d95cdfb368554bd6faec472e6089b802R72) where I had doubt, please check.

Then we should rethink returning an error if the token is empty. This won't work.

cc/ @baichinger

pkg/controller/oneagent/util_test.go
```diff
@@ -62,6 +62,43 @@ func TestOneAgent_Validate(t *testing.T) {
...
+		data["test_token_2"] = []byte("\t\n   dynatrace_test_token_2")
+		secret := corev1.Secret{Data: data}
+		token, err := getToken(&secret, "test_token")
+		token2, err := getToken(&secret, "test_token_2")
+
+		assert.NoError(t, err)
+		assert.Equal(t, token, "dynatrace_test_token")
+		assert.Equal(t, token2, "dynatrace_test_token_2")
+	}
+}
```
lrgar: Nitpicking: new line between functions.

PR #80, opened 2019-03-19
Properly watch for OneAgent pods recreation

lrgar:
The client.List() function falls back to a cache implementation after the initial attempt.

However, the client's cached implementation fails to handle complex queries, so rather than the function falling back to use the normal client, it just complains that the query is not supported.

On this commit, we query all the pods, and filter them ourselves as a workaround.

### DIFF
--- a/pkg/controller/oneagent/oneagent_controller.go
+++ b/pkg/controller/oneagent/oneagent_controller.go
@@ -14,7 +14,6 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
@@ -262,8 +261,10 @@ func (r *ReconcileOneAgent) reconcileVersion(reqLogger logr.Logger, instance *dy
 		instance.Status.Items = instances
 	}
 
+	reqLogger.Info("pods to delete", "count", len(podsToDelete))
+
 	// restart daemonset
-	err = r.deletePods(instance, podsToDelete)
+	err = r.deletePods(reqLogger, instance, podsToDelete)
 	if err != nil {
 		reqLogger.Error(err, "failed to update version")
 		return updateCR, err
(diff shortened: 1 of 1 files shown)


pkg/controller/oneagent/oneagent_controller.go
```diff
@@ -363,18 +364,24 @@ func newPodSpecForCR(instance *dynatracev1alpha1.OneAgent) corev1.PodSpec {
...
 		// delete pod
 		err := r.client.Delete(context.TODO(), &pod)
 		if err != nil {
 			return err
 		}
 
+		reqLogger.Info("waiting until pod is ready", "pod", pod.Name)
+
 		// wait for pod on node to get "Running" again
 		if err := r.waitPodReadyState(instance, pod); err != nil {
```
esukram: I know the `err` is print after this function is called, but at least indicating, that an error while being in the delete loop happened, could help.
What do you think?

pkg/controller/oneagent/oneagent_controller.go
```diff
@@ -363,18 +364,24 @@ func newPodSpecForCR(instance *dynatracev1alpha1.OneAgent) corev1.PodSpec {
...
 			return err
```
esukram: Is it intentional, that an error for a `pod` happens, the entire function is exited?
Breaking the loop and continuing with the remaining ones could still be valid.

pkg/controller/oneagent/oneagent_controller.go
```diff
@@ -383,22 +390,40 @@ func (r *ReconcileOneAgent) deletePods(instance *dynatracev1alpha1.OneAgent, pod
...
+			p := &podList.Items[i]
+			if p.Spec.NodeName != pod.Spec.NodeName || p.Status.Phase != corev1.PodRunning ||
+				p.ObjectMeta.Name == pod.Name {
+				continue
+			}
+			foundPods = append(foundPods, p)
+		}
+
+		if n := len(foundPods); n == 0 {
+			status = fmt.Errorf("waiting for pod to be recreated")
```
esukram: Please be more specific and descriptive.

pkg/controller/oneagent/oneagent_controller.go
lrgar: Yeah, it makes sense to update what we can.

pkg/controller/oneagent/oneagent_controller.go
lrgar: If we're going to try to update all the nodes rather than exiting quickly, then yes, this is needed to show errors per pod.

lrgar:
Hi @esukram - as mentioned, I agree that logging should be improved. Still, would you be OK with approving this PR, while I create another one with the logging changes?

lrgar:
Included changes to the log entries to include the node. Larger changes to be handled in a new PR.

pkg/controller/oneagent/oneagent_controller.go
```diff
@@ -366,22 +366,22 @@ func newPodSpecForCR(instance *dynatracev1alpha1.OneAgent) corev1.PodSpec {
...
+		reqLogger.Info("deleting pod", "pod", pod.Name, "node", pod.Spec.NodeName)
 
 		// delete pod
 		err := r.client.Delete(context.TODO(), &pod)
 		if err != nil {
 			return err
 		}
 
-		reqLogger.Info("waiting until pod is ready", "pod", pod.Name)
+		reqLogger.Info("waiting until pod is ready in node", "node", pod.Spec.NodeName)
```
esukram: typo: '... on the ...' but as we want to change the logic anyways...

PR #81, opened 2019-03-19
change readiness probe in pod spec to handle timeout

namratachaudhary:
Resolves the issue where a client's machine was unresponsive due to `pgrep -f` hogging memory. Changes the K8s readiness probe to use `ps -C` in place of that. 

Difficult to test this locally, or via brute-force.

### DIFF
--- a/pkg/controller/oneagent/oneagent_controller.go
+++ b/pkg/controller/oneagent/oneagent_controller.go
@@ -326,7 +326,6 @@ func newPodSpecForCR(instance *dynatracev1alpha1.OneAgent) corev1.PodSpec {
 				Handler: corev1.Handler{
 					Exec: &corev1.ExecAction{
 						Command: []string{
-							"timeout", "-s", "KILL", "-t", "60", // timeout in case grep fails
 							"grep", "oneagentwatchdo", "/proc/*/stat",
 						},
 					},
--- a/pkg/controller/oneagent/oneagent_controller.go
+++ b/pkg/controller/oneagent/oneagent_controller.go
@@ -325,11 +325,15 @@ func newPodSpecForCR(instance *dynatracev1alpha1.OneAgent) corev1.PodSpec {
 			ReadinessProbe: &corev1.Probe{
 				Handler: corev1.Handler{
 					Exec: &corev1.ExecAction{
-						Command: []string{"ps", "-C", "oneagentwatchdog"},
+						Command: []string{
+							"timeout", "-s", "KILL", "-t", "60", // timeout in case grep fails
+							"grep", "oneagentwatchdo", "/proc/*/stat",
+						},
 					},
 				},
 				InitialDelaySeconds: 30,
 				PeriodSeconds:       30,
+				TimeoutSeconds:      1,
 			},
 			Resources: instance.Spec.Resources,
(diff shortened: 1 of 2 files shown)


pkg/controller/oneagent/oneagent_controller.go
```diff
@@ -325,7 +325,7 @@ func newPodSpecForCR(instance *dynatracev1alpha1.OneAgent) corev1.PodSpec {
 			ReadinessProbe: &corev1.Probe{
 				Handler: corev1.Handler{
 					Exec: &corev1.ExecAction{
-						Command: []string{"pgrep", "-f", "oneagentwatchdog"},
+						Command: []string{"ps", "-C", "oneagentwatchdog"},
```
esukram: Did we ensure the OneAgent docker image contains `ps` and is able to handle the `-C` properly?

By properly I mean, the `-C` used to limit to 15 chars, and the command is 16 chars long:
>        -C cmdlist
>              Select by command name.  This selects the processes whose
>              executable name is given in cmdlist.  NOTE: The command name
>              is not the same as the command line. Previous versions of
>              procps and the kernel truncated this command name to 15
>              characters. This limitation is no longer present in both. If
>              you depended on matching only 15 characters, you may no longer
>              get a match.
http://man7.org/linux/man-pages/man1/ps.1.html

namratachaudhary: Typically, it should be given as `ps -C oneagentwatchdo`

namratachaudhary: Imo, we should add `apt-get update && apt-get install procps` while creating docker container for the same, there are several instances where ps does not work without explicit installation. 

Second, I'm still trying to see the advantage of using `ps` over `pgrep` - because if we're only concerned with the executable, then `pgrep <executable>` should work okay.

namratachaudhary: To handle installation of procps, we could add it to - `docker.io/dynatrace/oneagent:latest` perhaps?

pkg/controller/oneagent/oneagent_controller.go
```diff
@@ -325,11 +325,15 @@ func newPodSpecForCR(instance *dynatracev1alpha1.OneAgent) corev1.PodSpec {
 			ReadinessProbe: &corev1.Probe{
 				Handler: corev1.Handler{
 					Exec: &corev1.ExecAction{
-						Command: []string{"pgrep", "-f", "oneagentwatchdog"},
+						Command: []string{
+							"timeout", "-s", "KILL", "-t", "60", // timeout in case grep fails
```
baichinger: 60 seconds is way to high compared to the timeout (1) provided by handler. As there is already a timeout set in the handler I suggest to remove this one.

namratachaudhary: This command might not work on some docker images. e.g. it fails on openshift and ubuntu.

Also, as mentioned on Jira ticket, timeout (1) provided by handler does not "always" work, and leaves the thread hanging.

namratachaudhary:
Removed the timeout from exec of readiness probe. Currently relying on Timeout (1) set by handler. 
This was set as 1 by default, but there have been instances where it does not kill the process.

PR #82, opened 2019-03-21
Verifies istio installation in watch namespace

namratachaudhary:


namratachaudhary:
@lrgar could you check if this is what the expected behavior and result is?

pkg/controller/istio/helper.go
```diff
@@ -3,27 +3,58 @@ package istio
...
+
+	// Creates the dynamic interface.
+	dynamicClient, err := dynamic.NewForConfig(cfg)
+	if err != nil {
+		return err
+	}
+
+	namespace := os.Getenv(k8sutil.WatchNamespaceEnvVar)
+	//  List all of the Virtual Services.
+	virtualServices, err := dynamicClient.Resource(schema.GroupVersionResource{
```
esukram: We should place the resource also as static definition in `var` block of line `#18`.

pkg/controller/istio/helper.go
```diff
@@ -3,27 +3,58 @@ package istio
...
+		Group:   "networking.istio.io",
+		Version: "v1alpha3",
+	}).Namespace(namespace).List(metav1.ListOptions{})
```
esukram: Have we tested the retrieval of the `CRD` for the configured `namespace`?

pkg/controller/istio/helper.go
```diff
@@ -3,27 +3,58 @@ package istio
...
+	if err != nil {
+		return err
+	} else if len(virtualServices.Items) == 0 {
+		// no error, but no items either
+		return fmt.Errorf("no services found with group -  networking.istio.io in namespace %v", namespace)
```
esukram: I do not consider this as an error state, there is just no service defined - yet.

pkg/controller/oneagent/istio.go
```diff
@@ -13,13 +14,23 @@ import (
...
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 )
 
 func (r *ReconcileOneAgent) reconcileIstio(logger logr.Logger, instance *dynatracev1alpha1.OneAgent, dtc dtclient.Client) error {
 	var err error
 
-	// TODO determine if cluster runs istio
+	cfg, err := config.GetConfig()
+	if err != nil {
+		return err
```
esukram: The `err` should be logged.

pkg/controller/istio/helper.go
namratachaudhary: Yes, works as defined in the function.

pkg/controller/istio/helper.go
namratachaudhary: Yes.

pkg/controller/istio/helper.go
namratachaudhary: Yes

pkg/controller/istio/helper.go
```diff
@@ -3,27 +3,52 @@ package istio
...
 		Kind:    "ServiceEntry",
 	}
+
+	istioGVRName = "networking.istio.io"
 )
 
+// CheckIstioEnabled - Checks if Istio is installed
+func CheckIstioEnabled(cfg *rest.Config) (bool, error) {
+
+	client := discovery.NewDiscoveryClientForConfigOrDie(cfg)
```
lrgar: Don't kill the process because of this. `discovery.NewDiscoveryClientForConfig()` works here. Just return the error if it fails.

pkg/controller/istio/helper.go
```diff
@@ -3,27 +3,52 @@ package istio
...
+	apiGroupList, err := client.ServerGroups()
+	if err != nil {
+		return false, err
+	}
+
+	for _, apiGroup := range apiGroupList.Groups {
+
```
lrgar: Nitpicking: remove empty line.

pkg/controller/istio/helper.go
```diff
@@ -3,27 +3,52 @@ package istio
...
 		Version: "v1alpha3",
 		Kind:    "ServiceEntry",
 	}
+
+	istioGVRName = "networking.istio.io"
 )
 
+// CheckIstioEnabled - Checks if Istio is installed
+func CheckIstioEnabled(cfg *rest.Config) (bool, error) {
+
```
lrgar: Nitpicking: remove empty line.

pkg/controller/oneagent/istio.go
```diff
@@ -13,13 +14,24 @@ import (
...
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/config"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 )
 
 func (r *ReconcileOneAgent) reconcileIstio(logger logr.Logger, instance *dynatracev1alpha1.OneAgent, dtc dtclient.Client) error {
 	var err error
 
-	// TODO determine if cluster runs istio
+	cfg, err := config.GetConfig()
```
lrgar: When the ReconcileOneAgent object is created on newReconciler(), the Manager provides some objects. Get the config from there, e.g., `mgr.GetConfig()`

pkg/controller/oneagent/istio.go
```diff
@@ -13,13 +14,26 @@ import (
...
+	if err != nil {
+		logger.Error(err, "error getting default config")
+		return err
+	}
+
+	// determine if cluster runs istio in default cluster
+	exists, err = istio.CheckIstioEnabled(cfg)
```
lrgar: Missing colon `:=`. I'd also rename it to "enabled".

pkg/controller/oneagent/istio.go
```diff
@@ -13,13 +14,26 @@ import (
...
+	if err != nil {
+		logger.Error(err, "error checking for istio enabled")
+		return err
+	}
+	logger.Info(fmt.Sprintf("istio enabled : %v", exists))
```
lrgar: Something like `logger.Info("looked for Istio", "enabled", exists)`

namratachaudhary: I think there needs to be a case handling here, will double check with markus and implement.

pkg/controller/istio/helper.go
```diff
@@ -3,27 +3,53 @@ package istio
...
+	"github.com/operator-framework/operator-sdk/pkg/k8sutil"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/discovery"
+	"k8s.io/client-go/rest"
 )
 
 var (
+	// VirtualServiceGVK => definition of virtual service GVK for oneagent
 	VirtualServiceGVK = schema.GroupVersionKind{
 		Group:   "networking.istio.io",
```
esukram: Please use `istioGVRName` const in group names, to reduce code-duplication.

PR #84, opened 2019-03-22
Fix deploy for master

esukram:
Deployment manifest contains a new `subresources` definition, breaking the deployment.

### DIFF
--- a/deploy/kubernetes.yaml
+++ b/deploy/kubernetes.yaml
@@ -210,8 +210,6 @@ spec:
     singular: oneagent
   scope: Namespaced
   version: v1alpha1
-  subresources:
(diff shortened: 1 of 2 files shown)


lrgar:
The Status subresource was added by the Operator SDK 0.4.0.

To my understanding, it's required for enabling the status subresource for a CRD. See the CustomResourceDefinition snippet here: https://kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definitions/#subresources

I think this was new to K8s 1.10 though, but not sure what the behavior is without it.

esukram:
The manifest is correct, the original issue, seems to be different.

PR #85, opened 2019-03-25
Specifying shell for running readiness probe command

namratachaudhary:


pkg/controller/oneagent/oneagent_controller.go
```diff
@@ -345,7 +345,7 @@ func newPodSpecForCR(instance *dynatracev1alpha1.OneAgent) corev1.PodSpec {
 				Handler: corev1.Handler{
 					Exec: &corev1.ExecAction{
 						Command: []string{
-							"grep", "oneagentwatchdo", "/proc/*/stat",
+							"/bin/sh", "-c", "grep", "oneagent", "/proc/*/stat",
```
esukram: Why did we leave out the `oneagentwatchdog` and using plain `oneagent`?

namratachaudhary: My bad, copied the command I was using to test.

pkg/controller/oneagent/oneagent_controller.go
```diff
@@ -345,7 +345,7 @@ func newPodSpecForCR(instance *dynatracev1alpha1.OneAgent) corev1.PodSpec {
 				Handler: corev1.Handler{
 					Exec: &corev1.ExecAction{
 						Command: []string{
-							"grep", "oneagentwatchdo", "/proc/*/stat",
+							"/bin/sh", "-c", "grep", "oneagentwatchdog", "/proc/*/stat",
```
lrgar: Shouldn't the final command be?

```/bin/sh -c "grep oneagentwatchdo /proc/*/stat"```

i.e., three parameters rather than five: ```"/bin/sh", "-c", "grep oneagentwatchdo /proc/*/stat"```

Also note, oneagentwatchdog -> oneagentwatchdo. Stat files only have up to 15 chars for the process name.

PR #91, opened 2019-03-29
Backport feature to disable auto-updates

lrgar:


README.md
```diff
@@ -89,6 +89,8 @@ spec:
   #  limits:
   #    cpu: 300m
   #    memory: 1.5Gi
+  # disables automatic restarts of oneagent pods in case a new version is available
+  #disableAgentUpdate: true
```
esukram: We should state the default here, so is should state `false`.

deploy/cr.yaml
```diff
@@ -37,3 +37,5 @@ spec:
   #  limits:
   #    cpu: 300m
   #    memory: 1.5Gi
+  # disables automatic restarts of oneagent pods in case a new version is available
+  #disableAgentUpdate: true
```
esukram: We should state the default here, so is should state `false`.

lrgar:
Updated CR examples with the default disableAgentUpdate value. Will create another PR for the master branch, who has the same example.

PR #93, opened 2019-04-01
add unit test for one agent controller

namratachaudhary:


pkg/controller/oneagent/oneagent_controller_test.go
```diff
@@ -0,0 +1,86 @@
...
+
+func TestOneAgentController(t *testing.T) {
+
+	var (
+		name      = "dynatrace-oneagent"
+		namespace = "dynatrace"
+	)
+
+	oa := newOneAgentSpec()
+	dynatracev1alpha1.SetDefaults_OneAgentSpec(oa)
```
lrgar: Set the custom values (ApiUrl, and Tokens) before setting the defaults. That is the logic used with the customer.

pkg/controller/oneagent/oneagent_controller_test.go
```diff
@@ -0,0 +1,86 @@
...
+		Spec: *oa,
+	}
+	objs := []runtime.Object{
+		instance,
+	}
+
+	scheme := scheme.Scheme
+	scheme.AddKnownTypes(dynatracev1alpha1.SchemeGroupVersion, instance)
+
+	client := fake.NewFakeClient(objs...)
```
lrgar: You can pass instance, directly, right? e.g. `fake.NewFakeClient(instance)` so we don't need the `objs` variable.

pkg/controller/oneagent/oneagent_controller_test.go
```diff
@@ -0,0 +1,86 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/kubernetes/scheme"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+)
+
+func TestOneAgentController(t *testing.T) {
+
```
lrgar: Nitpicking: remove empty line.

pkg/controller/oneagent/oneagent_controller_test.go
```diff
@@ -0,0 +1,86 @@
...
+		namespace = "dynatrace"
+	)
+
+	oa := newOneAgentSpec()
+	dynatracev1alpha1.SetDefaults_OneAgentSpec(oa)
+	if oa.ApiUrl == "" {
+		oa.ApiUrl = "https://ENVIRONMENTID.live.dynatrace.com/api"
+	}
+	oa.Tokens = "token123"
+	ds := newDaemonSetSpec()
```
lrgar: The Reconciler defines the DaemonSet spec itself if it's not there. If we are testing how it behaves on a new environment, then I suggest leaving this task to the Reconciler.

pkg/controller/oneagent/oneagent_controller_test.go
```diff
@@ -0,0 +1,86 @@
...
+			Name:      name,
+			Namespace: namespace,
+		},
+	}
+	_, err := reconcileOA.Reconcile(req)
+	if err != nil {
+		t.Fatalf("error reconciling : %v", err)
+	}
+
+	// Check if deamonset has been created and has the correct size.
```
lrgar: "correct size" as in number of nodes?

pkg/controller/oneagent/oneagent_controller_test.go
```diff
@@ -0,0 +1,86 @@
...
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/kubernetes/scheme"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+)
+
+func TestOneAgentController(t *testing.T) {
```
lrgar: Since the test is quite long, a descriptive name would be helpful, e.g. ```TestReconcileOneAgent_ReconcileOnEmptyEnvironment```

PR #94, opened 2019-04-02
add workflow to deploy release snapshot

namratachaudhary:


.circleci/config.yml
```diff
@@ -67,6 +67,18 @@ jobs:
...
+    <<: *defaults
+    steps:
+      - setup_remote_docker
+      - run: |
+          mkdir -p ~/.docker && chmod 0700 ~/.docker
+          touch ~/.docker/config.json && chmod 0600 ~/.docker/config.json
+          base64 -d >~/.docker/config.json <<<"$OAO_DOCKER_AUTH"
+      - attach_workspace:
+          at: .
+      - run: docker build . -f ./build/Dockerfile -t $OAO_IMAGE:$CIRCLE_TAG
```
lrgar: We'd like to have a different tag for these and not just "release-0.2" Docker tags since it may confuse the customers. Something like "snapshot-$CIRCLE_TAG" perhaps.

.circleci/config.yml
```diff
@@ -67,6 +67,18 @@ jobs:
...
+      - run: docker push $OAO_IMAGE:$CIRCLE_TAG
```
lrgar: Same here.

esukram:
How do we deal with the latest snapshot in here?

So we should end up with:
* `latest` -> v0.3 -> v0.3.1
* `snapshot` -> `HEAD` of master branch
* `v0.3-snapshot` -> HEAD of `release-0.3` branch

Right?

.circleci/config.yml
esukram: As written above, I'd suggest:
`v0.3-snapshot` to conform with our release tags.

lrgar: Fine with me.

lrgar:
`latest` tag is created manually ATM AFAIK, which is why it's still 0.2.0

.circleci/config.yml
```diff
@@ -98,3 +110,9 @@ workflows:
               only: /^v.*/
             branches:
               ignore: /.*/
+      - deploy-release-snapshot:
+          requires:
+            - build
+          filters:
+            branches:
+              only: /release-*/
```
lrgar: It should be something like, `/^release-.*/`

.circleci/config.yml
```diff
@@ -67,6 +67,19 @@ jobs:
...
+    steps:
+      - setup_remote_docker
+      - run: |
+          mkdir -p ~/.docker && chmod 0700 ~/.docker
+          touch ~/.docker/config.json && chmod 0600 ~/.docker/config.json
+          base64 -d >~/.docker/config.json <<<"$OAO_DOCKER_AUTH"
+          export RELEASE_TAG=${CIRCLE_TAG//release-/v}-snapshot
+      - attach_workspace:
+          at: .
+      - run: docker build . -f ./build/Dockerfile -t $OAO_IMAGE:$RELEASE_TAG
```
lrgar: And as a final confirmation, are we sure that RELEASE_TAG is available here? Since they are two different `run` sections. Not sure how CircleCI handles this.

namratachaudhary: no it does not.
tested.

namratachaudhary: wrong env var, fixed now.

PR #98, opened 2019-04-09
add unit test for istio helper

namratachaudhary:


esukram:
I really like the contents of this PR!!

Furthermore the other case (not enabled) would be great to be there.

esukram:
Can we also add negative boundary tests like:
```go
ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
  w.WriteHeader(http.StatusUnauthorized)
}))
```

namratachaudhary:
> Can we also add negative boundary tests like:
> 
> ```go
> ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
>   w.WriteHeader(http.StatusUnauthorized)
> }))
> ```

Outside the scope of this PR. Kubernetes Discovery client does not account for permissions over API Resource list per se.

pkg/controller/istio/helper_test.go
```diff
@@ -0,0 +1,87 @@
...
+					},
+				},
+			},
+		}
+		server := initMockServer(t, list)
+		defer server.Close()
+		cfg := &restclient.Config{Host: server.URL}
+
+		r, e := CheckIstioEnabled(cfg)
+		if r != false {
```
lrgar: What about returned failures? i.e., `r` is false, but there is an error.

pkg/controller/istio/helper_test.go
```diff
@@ -0,0 +1,87 @@
...
+		defer server.Close()
+		cfg := &restclient.Config{Host: server.URL}
+
+		r, e := CheckIstioEnabled(cfg)
+		if r != true {
+			t.Fail()
+			t.Error(e)
+		}
+	}
+	{
```
lrgar: I'd prefer to have these two tests split on their own functions.

pkg/controller/istio/helper_test.go
```diff
@@ -0,0 +1,87 @@
...
+				},
+			},
+		}
+		server := initMockServer(t, list)
+		defer server.Close()
+		cfg := &restclient.Config{Host: server.URL}
+
+		r, e := CheckIstioEnabled(cfg)
+		if r != true {
+			t.Fail()
```
lrgar: `t.Error()` calls `t.Fail()` itself, so we don't need to call `t.Fail()` ourselves.

pkg/controller/istio/helper_test.go
```diff
@@ -0,0 +1,97 @@
...
+		t.Errorf("expected false, got true, %v", e)
+	}
+}
+
+func TestIstioWrongConfig(t *testing.T) {
+	// wrong config, we get error
+	list := &metav1.APIGroupList{}
+	server := initMockServer(t, list)
+	defer server.Close()
+	cfg := &restclient.Config{Host: "localhost:1000"}
```
lrgar: Is the "localhost:1000" the wrong config here? (We don't need the mock server then I think, right?)

pkg/controller/istio/helper_test.go
```diff
@@ -0,0 +1,97 @@
...
+				},
+			},
+		},
+	}
+	server := initMockServer(t, list)
+	defer server.Close()
+	cfg := &restclient.Config{Host: server.URL}
+
+	r, e := CheckIstioEnabled(cfg)
+	if r != true {
```
lrgar: Hmm, check for errors here as well. Same as the other tests.

pkg/controller/istio/helper_test.go
lrgar: And technically "localhost:1000" can be in use when running the tests locally I suppose.

namratachaudhary: No. It just means that wrong config has been provided, hence resulting in an error. Because otherwise we get no error, just "false"

pkg/controller/istio/helper_test.go
namratachaudhary: There is never a case where you get true and an error alongwith. It would be a redundant check.

pkg/controller/istio/helper_test.go
lrgar: What is the "wrong config" here, then?

Also, if having the server is correct then the host should be set to `server.URL`.

pkg/controller/istio/helper_test.go
lrgar: IMHO, I think it's a good habit to always check for errors first, regardless of the other values, but don't really mind in this case.

PR #100, opened 2019-04-18
change kubectl and oc commands to declarative commands

anmoel:
The preferred approach for managing resources is through declarative files called resource config used with the `kubectl apply` command. 
kubectl apply doesn't need to specify create or update, so you can use the same command for updating the dynatrace-oneagent-operator.

see https://kubectl.docs.kubernetes.io/


MartinGoodwell:
Thanks for the PR!
We'll have a look at your contribution and let you know

PR #112, opened 2019-05-03
Generate Istio objects for the API URL

lrgar:
On this PR we're now creating VirtualService and ServiceEntry objects for the API URL so that the Operator can access.

If Istio autoconfiguration is enabled, this happens before rolling the agent pods.

Also included, changing Istio log entries to have "istio:" prefix in an attempt to simplify the entries.

### DIFF
--- a/pkg/controller/oneagent/istio.go
+++ b/pkg/controller/oneagent/istio.go
@@ -17,45 +17,75 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 )
 
-func (r *ReconcileOneAgent) reconcileIstio(logger logr.Logger, instance *dynatracev1alpha1.OneAgent, dtc dtclient.Client) error {
+func (r *ReconcileOneAgent) reconcileIstio(logger logr.Logger, instance *dynatracev1alpha1.OneAgent, dtc dtclient.Client) (updated bool, ok bool) {
 	var err error
 
 	// Determine if cluster runs istio in default cluster
 	enabled, err := istio.CheckIstioEnabled(r.config)
 	if err != nil {
-		logger.Error(err, "error while checking for Istio availability")
-		return err
+		logger.Error(err, "istio: failed to verify Istio availability")
+		return false, false
 	}
 
-	logger.Info("Istio status", "enabled", enabled)
+	logger.Info("istio: status", "enabled", enabled)
 
 	if !enabled {
-		return nil
+		return false, true
+	}
+
+	apiHost, err := dtc.GetAPIURLHost()
+	if err != nil {
+		logger.Error(err, "istio: failed to get host for API URL")
+		return false, false
+	}
+
+	if upd, err := r.reconcileIstioConfigurations(logger, instance, []dtclient.CommunicationHost{apiHost}, "api-url"); err != nil {
+		logger.Error(err, "istio: error reconciling config for API URL")
+		return false, false
+	} else if upd {
+		return true, true
 	}
 
 	// Fetch endpoints via Dynatrace client
 	comHosts, err := dtc.GetCommunicationHosts()
 	if err != nil {
-		return err
+		logger.Error(err, "istio: failed to get communication endpoints")
+		return false, false
 	}
 
-	err = r.reconcileIstioCreateConfigurations(instance, comHosts, logger)
+	if upd, err := r.reconcileIstioConfigurations(logger, instance, comHosts, "communication-endpoint"); err != nil {
+		logger.Error(err, "istio: error reconciling config for communication endpoints")
+		return false, false
+	} else if upd {
+		return true, true
+	}
+
+	return false, true
+}
+
+func (r *ReconcileOneAgent) reconcileIstioConfigurations(logger logr.Logger, instance *dynatracev1alpha1.OneAgent,
+	comHosts []dtclient.CommunicationHost, role string) (bool, error) {
+	add, err := r.reconcileIstioCreateConfigurations(instance, comHosts, role, logger)
 	if err != nil {
-		logger.Error(err, "error reconciling Istio config")
-		return err
+		logger.Error(err, "istio: failed to create objects")
+		return false, err
 	}
 
-	r.reconcileIstioRemoveConfigurations(instance, comHosts, logger)
+	rem, err := r.reconcileIstioRemoveConfigurations(instance, comHosts, role, logger)
+	if err != nil {
+		logger.Error(err, "istio: failed to delete objects")
+		return false, err
+	}
 
-	return nil
+	return add || rem, nil
 }
 
 func (r *ReconcileOneAgent) reconcileIstioRemoveConfigurations(instance *dynatracev1alpha1.OneAgent,
-	comHosts []dtclient.CommunicationHost, logger logr.Logger) {
+	comHosts []dtclient.CommunicationHost, role string, logger logr.Logger) (bool, error) {
 
 	listOps := &client.ListOptions{
 		Namespace:     instance.Namespace,
-		LabelSelector: labels.SelectorFromSet(buildLabels(instance.Name)),
+		LabelSelector: labels.SelectorFromSet(buildIstioLabels(instance.Name, role)),
 	}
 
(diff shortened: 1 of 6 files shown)


pkg/controller/oneagent/istio.go
```diff
@@ -17,105 +17,152 @@ import (
...
+	logger.Info("istio: status", "enabled", enabled)
 
 	if !enabled {
-		return nil
+		return false, true
+	}
+
+	apiHost, err := dtc.GetAPIURLHost()
+	if err != nil {
+		logger.Error(err, "istio: failed to get host for API URL")
```
esukram: I would use `Dynatrace API` instead of plain `API`.

pkg/controller/oneagent/istio.go
```diff
@@ -17,105 +17,152 @@ import (
...
+		return false, false
+	}
+
+	if upd, err := r.reconcileIstioConfigurations(logger, instance, []dtclient.CommunicationHost{apiHost}, "api-url"); err != nil {
+		logger.Error(err, "istio: error reconciling config for API URL")
```
esukram: Same here.

pkg/controller/oneagent/istio.go
```diff
@@ -17,105 +17,152 @@ import (
...
+		return false, false
+	} else if upd {
+		return true, true
 	}
 
 	// Fetch endpoints via Dynatrace client
 	comHosts, err := dtc.GetCommunicationHosts()
 	if err != nil {
-		return err
+		logger.Error(err, "istio: failed to get communication endpoints")
```
esukram: Suggesting to use `Dynatrace communication endpoints` here.

pkg/controller/oneagent/istio.go
```diff
@@ -17,105 +17,152 @@ import (
...
+		return false, false
+	}
+
+	if upd, err := r.reconcileIstioConfigurations(logger, instance, comHosts, "communication-endpoint"); err != nil {
+		logger.Error(err, "istio: error reconciling config for communication endpoints")
```
esukram: Same here.

pkg/controller/oneagent/istio.go
```diff
@@ -17,105 +17,152 @@ import (
...
 	seen := map[string]bool{}
 	for _, ch := range comHosts {
 		seen[istio.BuildNameForEndpoint(instance.Name, ch.Host, ch.Port)] = true
 	}
 
-	r.reconcileIstioRemoveConfiguration(instance, istio.VirtualServiceGVK, listOps, seen, logger)
-	r.reconcileIstioRemoveConfiguration(instance, istio.ServiceEntryGVK, listOps, seen, logger)
+	vsUpd, err := r.reconcileIstioRemoveConfiguration(instance, istio.VirtualServiceGVK, listOps, seen, logger)
+	if err != nil {
+		return false, err
```
esukram: In case of any recurring failure during removing the `VS` objects, we never clean up the `SE` objects. Should we continue, even if an error occurred?

pkg/controller/oneagent/istio.go
```diff
@@ -17,105 +17,152 @@ import (
...
 
 	var list unstructured.UnstructuredList
 	list.SetGroupVersionKind(gvk)
 
 	if err := r.client.List(context.TODO(), listOps, &list); err != nil {
-		return
+		return false, err
 	}
 
+	upd := false
```
esukram: This var should be named `del` or similar, to not confuse with the *update* term.

pkg/controller/oneagent/istio.go
```diff
@@ -17,105 +17,152 @@ import (
...
 	}
+
+	return upd, nil
 }
 
 func (r *ReconcileOneAgent) reconcileIstioCreateConfigurations(instance *dynatracev1alpha1.OneAgent,
-	comHosts []dtclient.CommunicationHost, logger logr.Logger) error {
+	comHosts []dtclient.CommunicationHost, role string, logger logr.Logger) (bool, error) {
+
+	upd := false
```
esukram: Would it make sense to name this var also differently? Like `created` or similar...

pkg/controller/oneagent/istio.go
```diff
@@ -17,105 +17,152 @@ import (
...
 	for _, ch := range comHosts {
 		name := istio.BuildNameForEndpoint(instance.Name, ch.Host, ch.Port)
 
 		if notFound := r.configurationExists(istio.ServiceEntryGVK, instance.Namespace, name); notFound {
 			logger.Info("creating Istio ServiceEntry", "objectName", name, "host", ch.Host, "port", ch.Port)
 			payload := istio.BuildServiceEntry(name, ch.Host, ch.Port, ch.Protocol)
-			if err := r.reconcileIstioCreateConfiguration(instance, istio.ServiceEntryGVK, payload); err != nil {
-				logger.Info(fmt.Sprintf("failed to create Istio ServiceEntry: %v", err))
+			if err := r.reconcileIstioCreateConfiguration(instance, istio.ServiceEntryGVK, role, payload); err != nil {
+				return false, fmt.Errorf("failed to create Istio ServiceEntry: %v", err)
```
esukram: Is exiting the function here the right choice?

Logging the failure and continue could be also ok-ish.

pkg/controller/oneagent/istio.go
```diff
@@ -17,105 +17,152 @@ import (
...
+			upd = true
 		}
 
 		if notFound := r.configurationExists(istio.VirtualServiceGVK, instance.Namespace, name); notFound {
 			logger.Info("creating Istio VirtualService", "objectName", name, "host", ch.Host, "port", ch.Port, "protocol", ch.Protocol)
 			payload := istio.BuildVirtualService(name, ch.Host, ch.Port, ch.Protocol)
-			if err := r.reconcileIstioCreateConfiguration(instance, istio.VirtualServiceGVK, payload); err != nil {
-				logger.Info(fmt.Sprintf("failed to create Istio VirtualService: %v", err))
+			if err := r.reconcileIstioCreateConfiguration(instance, istio.VirtualServiceGVK, role, payload); err != nil {
+				return false, fmt.Errorf("failed to create Istio VirtualService: %v", err)
```
esukram: Is exiting the function here the right choice?

Logging the failure and continue could be also ok-ish.

pkg/controller/oneagent/istio.go
```diff
@@ -147,3 +194,9 @@ func (r *ReconcileOneAgent) configurationExists(gvk schema.GroupVersionKind, nam
 
 	return errors.IsNotFound(r.client.Get(context.TODO(), key, &objQuery))
 }
+
+func buildIstioLabels(name, role string) map[string]string {
```
esukram: Like that a lot!

pkg/dynatrace-client/client.go
```diff
@@ -368,3 +344,40 @@ func readCommunicationHosts(r io.Reader) ([]CommunicationHost, error) {
 
 	return out, nil
 }
+
+func parseEndpoint(s string) (CommunicationHost, error) {
```
esukram: Would love to see some unit tests for this new utility function ;)

pkg/controller/oneagent/istio.go
```diff
@@ -17,105 +17,152 @@ import (
...
 
 	if err := r.client.List(context.TODO(), listOps, &list); err != nil {
-		return
+		return false, err
 	}
 
+	upd := false
+
 	for _, item := range list.Items {
 		if _, ok := seen[item.GetName()]; !ok {
```
esukram: `ok` should be something more meaningful, like _removed_ or unseen ;)

lrgar: Will use `exists`: if it doesn't exists then...

lrgar:
Regarding the comments about exiting quickly vs logging and continue, there is nothing wrong with the latter aside from some extra logging. I assume they will mostly all succeed or fail rather than some in particular.

Anyway, this also needs tests to verify what the behavior with removals, but another PR is better for that.

esukram:
Yes, having changes scoped per PR makes perfect sense!

PR #117, opened 2019-06-12
add reconcile test for istio

namratachaudhary:


namratachaudhary:
The test fails because we are unable to query gvr for istio in given namespace.

lrgar:
I presume the error is caused because `getVirtualServices` is using the mock HTTP server, rather than the fake client, and the former returns 404 for almost everything.

namratachaudhary:
There is code duplication in some parts of tests. We need to work on a better framework to run unit tests on operator, separate out commonly used code. 
Maybe in a later issue some time.

PR #118, opened 2019-06-14
Mocking DT Api/client for Istio objects

namratachaudhary:


pkg/controller/oneagent/istio_test.go
```diff
@@ -129,6 +42,21 @@ func TestReconcileOneAgent_ReconcileIstio(t *testing.T) {
 	}
 }
 
+func TestReconcileOneAgent_ReconcileIstioViaDynatraceClient(t *testing.T) {
```
lrgar: The `TestReconcileOneAgent_ReconcileIstio` test above does everything done in this test as well. Wouldn't that make `TestReconcileOneAgent_ReconcileIstioViaDynatraceClient` superfluous?

pkg/controller/oneagent/oneagent_controller_test.go
```diff
@@ -2,27 +2,99 @@ package oneagent
...
+	os.Setenv(k8sutil.WatchNamespaceEnvVar, "dynatrace")
+
+	server := initMockServer(t)
+	defer server.Close()
 
 	oa := newOneAgentSpec()
-	oa.ApiUrl = "https://ENVIRONMENTID.live.dynatrace.com/api"
+	oa.ApiUrl = testAPIUrl
 	oa.Tokens = "token_test"
+	oa.EnableIstio = true
```
lrgar: Do we want to set `EnableIstio = true` for all tests? There is a test for basic reconciliation without checking for Istio behavior.

pkg/controller/oneagent/istio_test.go
namratachaudhary: `TestReconcileOneAgent_ReconcileIstioViaDynatraceClient` might need more work in terms of testing istio endpoints which can either be added in consequent PR.

PR #119, opened 2019-06-19
Change in communication endpoints should update istio config

namratachaudhary:


pkg/controller/oneagent/istio_test.go
```diff
@@ -68,6 +71,29 @@ func TestReconcileOneAgent_ReconcileIstioViaDynatraceClient(t *testing.T) {
...
+		Host:     "https://endpoint3.dev.ruxitlabs.com/communication",
+		Port:     443,
+	})
+
+	var log = logf.Log.WithName("oneagent.controller.test")
+
+	upd, err := reconcileOA.reconcileIstioConfigurations(log, instance, commHosts, "communication-endpoint")
+	if upd != true {
+		t.Error("expected true got false, communication endpoints needed to be updated")
+	}
```
lrgar: We're only checking whether the reconciliation succeeded or failed, right? I suggest checking at least for the number of ServiceEntry objects.

Please also add a check after removal of a communication endpoint.

namratachaudhary: There does not seem to be state management from istio objects per se. 
The only persistence is `seen := map[string]bool{}` which happens per-request and it flushed in consequent function call. Thus, `reconcileIstioRemoveConfigurations` returns true almost always.

Also, the function `reconcileIstioRemoveConfiguration` removes the entry if present and recreates it. Not skip or throw an exception. In which case, that function also always returns true. Even in case of error, it continues to the next item in the list after logging the error. 

All these functions are written to assert true at every step of the way, no matter the input. Please let me know how I can design the functional tests to output different results. 

Also, pointing out - function [reconcileIstio](https://github.com/Dynatrace/dynatrace-oneagent-operator/blob/master/pkg/controller/oneagent/istio.go#L42) always returns at this point. As you can see in the if-else statement. This stems from the fact that `reconcileIstioConfigurations` almost always returns true. I'm not sure if the corresponding workflow which follows beyond this point even gets executed, which is basically, resolving istio objects on communication endpoints obtained from dynatrace client. 

I will test this further and let you know what I find. But as per my understanding, there seems to be a lot of places where the code does not reach, at least in the test environment. Not sure how this is functioning in production.

namratachaudhary: Additionally, this function is responsible for finding out whether object config exists on k8s client or not
https://github.com/Dynatrace/dynatrace-oneagent-operator/blob/master/pkg/controller/oneagent/istio.go#L175

The client used here is the default k8s client, but for tests, we use the fake client. While this works in production because default k8s client, and the client associated with reconciler accesses same host and port, it might not be the case while writing unit tests, where the fake client and default k8s client do not share same set of resources.

lrgar: You should be able to query the created objects by, let's say, listing objects from `reconcileOA.client`, right? We can confirm the created objects if this is the case.

namratachaudhary: I get nothing. An empty list. Whereas, in the code flow, the objects are shown as created. querying them post reconcile workflow does not work at all :confused:

namratachaudhary:
Closing this as the code logic was incorrect, this has been attempted to be fixed in #121

PR #121, opened 2019-06-25
Using interface to query istio CRD

namratachaudhary:


pkg/controller/oneagent/istio.go
```diff
@@ -84,37 +102,79 @@ func (r *ReconcileOneAgent) reconcileIstioRemoveConfigurations(instance *dynatra
...
+func (r *ReconcileOneAgent) reconcileIstioRemoveConfiguration(
+	ic *versionedIstioClient.Clientset,
+	gvk schema.GroupVersionKind,
+	listOps *client.ListOptions,
+	seen map[string]bool,
+	logger logr.Logger) bool {
 
-	if err := r.client.List(context.TODO(), listOps, &list); err != nil {
-		logger.Error(err, fmt.Sprintf("istio: failed to list %s objects", gvk.Kind))
+	ic, err := r.initialiseIstioClient(logger)
```
lrgar: `ic` already comes from the function args.

pkg/controller/oneagent/istio.go
```diff
@@ -84,37 +102,79 @@ func (r *ReconcileOneAgent) reconcileIstioRemoveConfigurations(instance *dynatra
...
+	}
+	return ic, err
+}
 
-	var list unstructured.UnstructuredList
-	list.SetGroupVersionKind(gvk)
+func (r *ReconcileOneAgent) reconcileIstioRemoveConfiguration(
+	ic *versionedIstioClient.Clientset,
+	gvk schema.GroupVersionKind,
+	listOps *client.ListOptions,
```
lrgar: `listOps` is not used. IIRC, it had some labels to differentiate communication endpoints from different sets. Does this logic work?

pkg/controller/oneagent/istio.go
```diff
@@ -84,37 +102,79 @@ func (r *ReconcileOneAgent) reconcileIstioRemoveConfigurations(instance *dynatra
...
 
-	del := false
-
-	for _, item := range list.Items {
-		if _, inUse := seen[item.GetName()]; !inUse {
-			logger.Info(fmt.Sprintf("istio: removing %s: %v", gvk.Kind, item.GetName()))
-			if err := r.client.Delete(context.TODO(), &item); err != nil {
-				logger.Error(err, fmt.Sprintf("istio: failed to delete %s", gvk.Kind))
-				continue
+	if gvk == istio.ServiceEntryGVK {
```
lrgar: I'd suggest splitting the function by GVK since no logic is shared between the two if branches.

pkg/controller/oneagent/istio_test.go
```diff
@@ -1,16 +1,22 @@
...
 import (
-	"context"
+	"bytes"
+	"encoding/json"
+	"os"
 	"testing"
 
+	_ "github.com/Dynatrace/dynatrace-oneagent-operator/pkg/apis"
 	dynatracev1alpha1 "github.com/Dynatrace/dynatrace-oneagent-operator/pkg/apis/dynatrace/v1alpha1"
+	fakeIstio "github.com/Dynatrace/dynatrace-oneagent-operator/pkg/apis/networking/clientset/versioned/fake"
```
lrgar: Packages in lowercase.

pkg/controller/oneagent/istio_test.go
```diff
@@ -1,16 +1,22 @@
...
+	istioV1alpha3 "github.com/Dynatrace/dynatrace-oneagent-operator/pkg/apis/networking/istio/v1alpha3"
```
lrgar: Here too.

pkg/controller/oneagent/istio_test.go
```diff
@@ -19,34 +25,43 @@ var (
 	namespace  = "dynatrace"
 )
 
-func TestReconcileOneAgent_ReconcileIstio(t *testing.T) {
-	oa := newOneAgentSpec()
-	oa.ApiUrl = testAPIUrl
-	oa.Tokens = "token_test"
-	oa.EnableIstio = true
-	dynatracev1alpha1.SetDefaults_OneAgentSpec(oa)
+func TestReconcileOneAgent_CreateIstioObjects(t *testing.T) {
```
lrgar: Hmm, at the very least we aren't creating any objects here? At least not by the reconciler.

pkg/controller/oneagent/istio_test.go
```diff
@@ -19,34 +25,43 @@ var (
...
-	serviceEntry := getGVK(client, istio.ServiceEntryGVK)
-	if serviceEntry == nil {
-		t.Error("no istio objects for service entry")
+	if len(vsList.Items) == 0 {
+		t.Error("Expected items, got nil")
 	}
+	t.Logf("list of istio object %v", vsList.Items)
 }
 
 func TestReconcileOneAgent_ReconcileIstioViaDynatraceClient(t *testing.T) {
```
lrgar: Is it possible to check for actually created objects? We're only reconciling twice and checking whether if the function was successful or not, which doesn't say much.

namratachaudhary: Not without the actual mock flow for rest client. Can reserve it for housekeeping?

lrgar: OK, we can take it there.

pkg/apis/networking/istio/v1alpha3/service_entry.go
```diff
@@ -0,0 +1,87 @@
...
+}
+
+// ServiceEntrySpec is a wrapper around Istio ServiceEntry
+type ServiceEntrySpec struct {
+	istiov1alpha3.ServiceEntry
+}
+
+// DeepCopyInto is a deepcopy function, copying the receiver, writing into out. in must be non-nil.
+// Based of https://github.com/istio/istio/blob/release-0.8/pilot/pkg/config/kube/crd/types.go#L450
+func (in *ServiceEntrySpec) DeepCopyInto(out *ServiceEntrySpec) {
```
lrgar: This doesn't really deep-copy but don't know how important that is.

pkg/apis/networking/istio/v1alpha3/virtual_service.go
```diff
@@ -0,0 +1,88 @@
...
+
+// VirtualServiceSpec is a wrapper around Istio VirtualService
+type VirtualServiceSpec struct {
+	istiov1alpha3.VirtualService
+}
+
+// DeepCopyInto is a deepcopy function, copying the receiver, writing into out. in must be non-nil.
+// Based of https://github.com/istio/istio/blob/release-0.8/pilot/pkg/config/kube/crd/types.go#L450
+func (in *VirtualServiceSpec) DeepCopyInto(out *VirtualServiceSpec) {
+	*out = *in
```
lrgar: Same here.

namratachaudhary: This uses the deepcopy method of istio's Virtual Service, I think. Tried it with Spec and buffers. Check the tests?

lrgar: Hmm, I don't think Go can have the assignment operator overridden to use DeepCopy.

As far as I know, this instruction copies all of the fields from one object to another, including the references, but not the actual references' content. So modifying the value in some list from object, it also applies to the other. If you need actual DeepCopy you need to explicitly call something that does it.

But as I said, not sure how important is that to us.

PR #123, opened 2019-07-11
Watch nodes and send mark for termination event

namratachaudhary:
- [x] Add NodeWatcher

- [x] Decouple MockDynatraceClient 

- [x] Implement logic to post event

- [x] Unit Tests


pkg/controller/oneagent/oneagent_controller.go
```diff
@@ -245,6 +255,44 @@ func (r *ReconcileOneAgent) buildDynatraceClient(instance *dynatracev1alpha1.One
...
 }
 
+func (r *ReconcileOneAgent) reconcileNodesMarkedForDeletion(reqLogger logr.Logger, dtc dtclient.Client) error {
+
+	// nodes := &corev1.NodeList(metav1.ListOptions{})
+	nodeList := &corev1.NodeList{}
+	fieldSelector := fields.SelectorFromSet(fields.Set{
+		"spec.unschedulable": "True",
+	})
+	listOps := &client.ListOptions{
```
lrgar: The OneAgent object also has a node selector that should be included here. Customers may have multiple OneAgent objects with different node selectors (e.g., production and staging, etc.)

pkg/controller/oneagent/oneagent_controller.go
```diff
@@ -245,6 +255,44 @@ func (r *ReconcileOneAgent) buildDynatraceClient(instance *dynatracev1alpha1.One
...
+	if err != nil {
+		reqLogger.Error(err, "failed to list nodes", "listops", listOps)
+		return err
+	}
+
+	for _, node := range nodeList.Items {
+		if node.Spec.Unschedulable {
+			reported, ok := cordonedNodes[&node]
+			if !ok {
+				cordonedNodes[&node] = bool(false)
```
lrgar: Any reason for `bool(false)` rather than `false`?

pkg/controller/oneagent/oneagent_controller.go
```diff
@@ -245,6 +255,44 @@ func (r *ReconcileOneAgent) buildDynatraceClient(instance *dynatracev1alpha1.One
...
+	}
+	err := r.client.List(context.TODO(), listOps, nodeList)
+	if err != nil {
+		reqLogger.Error(err, "failed to list nodes", "listops", listOps)
+		return err
+	}
+
+	for _, node := range nodeList.Items {
+		if node.Spec.Unschedulable {
+			reported, ok := cordonedNodes[&node]
```
lrgar: Since the key is a pointer, different runs of the reconciler may have the same node with different pointers, can't they?

Also, assuming that the pointer remains the same, or we use the name as a key. The scenario: cordon node, uncordon node, cordon node, with the Operator never stopping in the meanwhile, we wouldn't receive the notification for the second cordoning, right?

pkg/dynatrace-client/client.go
```diff
@@ -206,6 +210,52 @@ func (c *client) makeRequest(format string, a ...interface{}) (*http.Response, e
...
+			  ]
+			}
+		  ]
+		},
+		"source": "OpsControl",
+		"annotationType": "defect",
+		"annotationDescription": "The node is marked for termination"
+	  }
+	`
+	bbytes := []byte(fmt.Sprintf(body, time.Now().Unix(), nodeID))
```
lrgar: Do we need to send here the node name, or the Dynatrace's host ID used? If it's the latter, then you may need to query the host ID from the IP. We do something similar for getting the agent version on the host.

pkg/dynatrace-client/client.go
```diff
@@ -206,6 +210,52 @@ func (c *client) makeRequest(format string, a ...interface{}) (*http.Response, e
 	return c.httpClient.Get(url)
 }
 
+// PostMarkedForTerminationEvent =>
+// send event to dynatrace api that an event has been marked for termination
+func (c *client) PostMarkedForTerminationEvent(nodeID string) (string, error) {
+
+	url := fmt.Sprintf("%s/v1events", c.url)
```
lrgar: Is this URL OK? Perhaps, `/v1/events`?

pkg/controller/oneagent/oneagent_controller.go
namratachaudhary: Node selector for selecting nodes? I dont think the nodes are controller objects, and the only relevance is them being cordoned.

lrgar: A given OneAgent object has a node selector, which says which nodes will have the OneAgent pod running. It doesn't necessarily have to be all nodes in the cluster.

You may receive the cordoning notification about a node not being monitored by the OneAgent and yet still trying to send it.

pkg/dynatrace-client/client.go
namratachaudhary: The doc said, nodeID

lrgar: Whatever it's needed, but I'd be really surprised if it's not the Dynatrace Host Entity ID.

pkg/controller/oneagent/oneagent_controller.go
```diff
@@ -245,6 +255,60 @@ func (r *ReconcileOneAgent) buildDynatraceClient(instance *dynatracev1alpha1.One
...
+					return err
+				}
+				reqLogger.Info("reconcileNodesMarkedForDeletion: event sent, status %s", status)
+
+				// cordoned, in map, and reported
+				cordonedNodes[nodeInternalIP] = true
+			}
+		}
+		// if uncordoned, but present in map => remove from map
+		if !cordoned && ok {
```
lrgar: Hmm, since we are filtering by `fieldSelector: spec.unschedulable=true`, are we going to get any uncordoned nodes?

pkg/controller/oneagent/oneagent_controller.go
```diff
@@ -245,6 +255,60 @@ func (r *ReconcileOneAgent) buildDynatraceClient(instance *dynatracev1alpha1.One
 	return dtc, err
 }
 
+func (r *ReconcileOneAgent) reconcileNodesMarkedForDeletion(
+	reqLogger logr.Logger,
+	instance *dynatracev1alpha1.OneAgent,
+	dtc dtclient.Client) error {
+
+	// nodes := &corev1.NodeList(metav1.ListOptions{})
```
lrgar: Nitpicking: remove unused comments.

pkg/controller/oneagent/oneagent_controller.go
```diff
@@ -245,6 +255,60 @@ func (r *ReconcileOneAgent) buildDynatraceClient(instance *dynatracev1alpha1.One
...
+	nodeList := &corev1.NodeList{}
+	fieldSelector := fields.SelectorFromSet(fields.Set{
+		"spec.unschedulable": "True",
+	})
+	labelSelector := labels.SelectorFromSet(labels.Set(instance.Spec.NodeSelector))
+	listOps := &client.ListOptions{
+		//--field-selector=spec.unschedulable=True
```
lrgar: Same here.

pkg/controller/oneagent/oneagent_controller.go
namratachaudhary: Should we remove the field selector and iterate over all the nodes instead?

lrgar: Yes, I think that is the way to go. We already go through every node to look for the agent versions anyway.

PR #125, opened 2019-07-16
Host info map

namratachaudhary:


pkg/dynatrace-client/client_test.go
```diff
@@ -155,19 +155,18 @@ const (
 )
 
 func TestReadHostMap(t *testing.T) {
```
lrgar: Can we add the `entityId` entry to `goodHostsResponse` to confirm that everything is good with the parsing?

PR #127, opened 2019-08-02
Use Kubebuilder and envtest.Environment to test the reconciler operations

lrgar:
The controller-runtime library, on which the Operator Framework SDK is based on, [recommends using `envtest.Environment`](https://github.com/kubernetes-sigs/controller-runtime/blob/ee41a803ccc9c8a308b751433fc35282c716a9ff/FAQ.md#q-how-should-i-write-tests--any-suggestions-for-getting-started) class to test the behavior of controllers (and operators). This class depends on Kubebuilder, who provides a real instance of the API Server.

Rather than using mocks, the tests have been adapted here to use such envtest.Environment class. For example, we can now query Istio objects from this API Server instance to confirm whether the Operator has created or deleted them as expected. I've also added an additional class to prepare a configured reconciler instance through `newTestEnvironment` which we can use for unit testing, as shown in some tests.

Since we don't need to mock the client anymore, for the generated Istio types, we can remove the Clientset implementation and some others. I'll include that in a follow-up PR.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -61,6 +61,9 @@ jobs:
       # Run tests and build the binaries.
       - run:
           name: Run tests
+          command: go test -cover -v $(go list ./... | circleci tests split)
+      - run:
+          name: Run integration tests
           command: go test -cover -tags integration -v $(go list ./... | circleci tests split)
       - run:
           name: Build binary
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -61,7 +61,7 @@ jobs:
       # Run tests and build the binaries.
       - run:
           name: Run tests
-          command: go test -cover -v $(go list ./... | circleci tests split)
+          command: go test -cover -tags integration -v $(go list ./... | circleci tests split)
       - run:
           name: Build binary
           command: go build -o ./build/_output/bin/dynatrace-oneagent-operator ./cmd/manager
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -20,6 +20,30 @@ jobs:
     parallelism: 4
     steps:
       - checkout
+
+      # Download and cache Kubebuilder.
+      - run:
+          name: Prepare directories for Kubebuilder
+          command: |
+            sudo mkdir -p /usr/local/kubebuilder
+            sudo chmod 777 /usr/local/kubebuilder
+      - restore_cache:
+          key: kubebuilder-1.0.8
+          paths:
+              - /usr/local/kubebuilder/bin
+      - run:
+          name: Fetch Kubebuilder
+          command: |
+            if [ ! -d "/usr/local/kubebuilder/bin" ]; then
+              curl -L https://github.com/kubernetes-sigs/kubebuilder/releases/download/v1.0.8/kubebuilder_1.0.8_linux_amd64.tar.gz -o kubebuilder.tar.gz
+              tar -zxvf kubebuilder.tar.gz --strip-components=1 -C /usr/local/kubebuilder
+            fi
+      - save_cache:
+          key: kubebuilder-1.0.8
+          paths:
+              - /usr/local/kubebuilder/bin
+
+      # Download an cache dependencies.
       - restore_cache:
           key: gopkg-{{ checksum "Gopkg.toml" }}
           paths:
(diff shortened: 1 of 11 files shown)


pkg/controller/oneagent/envutils_test.go
```diff
@@ -0,0 +1,199 @@
...
+	}); err != nil {
+		svr.Stop()
+		return nil, err
+	}
+
+	e := &ControllerTestEnvironment{
+		server: svr,
+		Client: c,
+		CommunicationHosts: []string{
+			"https://endpoint1.dev.ruxitlabs.com/communication",
```
esukram: Please get rid of `ruxitlabs` and use `dynatracelabs` instead.

pkg/controller/oneagent/envutils_test.go
```diff
@@ -0,0 +1,199 @@
...
+			"https://endpoint2.dev.ruxitlabs.com/communication",
```
esukram: Same here.

esukram:
> Additionally, [kubebuilder](https://github.com/kubernetes-sigs/kubebuilder) is mainly used for building kubernetes apis for CRDs, and test it against an existing cluster. This is similar to operator's partial functionality of reconcile flow, but not entirely.. since operator-sdk acts on the principle of lifecycle management. So basically this PR brings further complexity of combining two operator frameworks in one project.

Once we use `kubebuilder` doesn't the whole tests behave more like integration tests? Thus we should split the test suite into two different types:

1. Unit-tests, fast, have no external dependencies
2. Integration-tests, using APIs and processes - I just stumbled across this [SO](https://stackoverflow.com/a/28007631)

namratachaudhary:
@esukram You basically reiterated my first concern :grinning: 
Using kubebuilder would make these as integration tests. I still believe, that integration tests can be run against an existent kubernetes api-server, and would not require setting up kubebuilder additionally. 

For writing unit-tests, which in current design seem tricky to implement (due to lack of use of interfaces and tightly coupled components in controller), we can take a look at refactoring certain components which will make it easier to extend operability of the controller and testing them as units instead of integrations.

namratachaudhary:
As discussed, please annotate the tests so that they signify integration tests. In the later versions, we can take a look at abstracting out components to facilitate unit testing. 

Also, in the meantime, if it is not entirely necessary, do hold off on removing clientsets as we would most certainly be using them in the future.

pkg/controller/oneagent/envutils_test.go
lrgar: The value itself is not really relevant, changed to `https://endpoint1.test.com/communication`

lrgar:
Finally got time to get back to this.

Yes, these are integration tests. I was able to run them as unit tests, but following with controller-runtime's documentation recommendation I went with `envtest.Environment` (i.e., integration/feature tests.) I was on the fence on whether use build flags (people tend to not run them otherwise) but don't mind adding them. I've added a commit with this change.

Also for clarification, the Operator SDK is based on `controller-runtime`, which is a subcomponent of Kubebuilder. The `envtest.Environment` class then uses the `kube-apiserver` and `etcd` to simulate a Kubernetes environment. The Kubebuilder package includes these two binaries inside along the `kubebuilder` CLI itself, but we can ignore the last one.

.circleci/config.yml
```diff
@@ -61,7 +61,7 @@ jobs:
       # Run tests and build the binaries.
       - run:
           name: Run tests
-          command: go test -cover -v $(go list ./... | circleci tests split)
+          command: go test -cover -tags integration -v $(go list ./... | circleci tests split)
```
namratachaudhary: would be good to run tests separately - integration and unit tests
i dont know what this change exactly does.

.circleci/config.yml
```diff
@@ -20,6 +20,30 @@ jobs:
     parallelism: 4
     steps:
       - checkout
+
+      # Download and cache Kubebuilder.
+      - run:
```
namratachaudhary: can we separate out these steps to run as integration test job instead of overwriting the existing workflow?

.circleci/config.yml
lrgar: To both comments, I'm concerned about perhaps losing the coverage data, don't know if it gets overridden. `-tags integration` includes unit tests either way.

Are we using publishing the coverage output somewhere ATM?

lrgar: Spoke with Namrata. Integration steps split to their own step.

.circleci/config.yml
lrgar: It's possible, but would require a bigger change on the structure of the workflows. Don't mind doing it later.

PR #137, opened 2019-09-12
Handle scenario when Istio is installed after the Operator is started

lrgar:
Issues handled in this PR - 
* Infinite reconciliation loop while saving CR status when no oneagent pods are running
* Operator needs a restart when istio is deployed after installing oneagent.

Fixes - 
* Infinite reconciliation loop is handled by comparing instances as long as there is data and handling - with a buffer of 30 minutes before requeue request.
* Operator needs to be restarted because RestMapper does not update the new installed CRDs. This issue is open and tracked in [PR on controller-runtime](https://github.com/kubernetes-sigs/controller-runtime/issues/321).

### DIFF
--- a/README.md
+++ b/README.md
@@ -155,7 +155,7 @@ $ oc delete -n dynatrace oneagent --all
 $ oc delete -f https://raw.githubusercontent.com/Dynatrace/dynatrace-oneagent-operator/master/deploy/openshift.yaml
 ```
 
-##### Known Limitation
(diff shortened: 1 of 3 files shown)


namratachaudhary:
Please detail out what was breaking, why and how does this PR fixes it. 
Thanks.

lrgar:
I added explanations to the commits themselves, I can explain further if necessary.

PR #142, opened 2019-09-27
Fix panic in Istio reconciler

baichinger:
- Fixed panic in reconciler for IP based entries.
- Fixed log spam.
- Refactored reconciliation logic into smaller methods.


pkg/controller/oneagent/istio.go
```diff
@@ -181,65 +191,93 @@ func (r *ReconcileOneAgent) removeIstioConfigurationForVirtualService(
 	return del
 }
 
-func (r *ReconcileOneAgent) reconcileIstioCreateConfigurations(
-	instance *dynatracev1alpha1.OneAgent,
-	ic *versionedistioclient.Clientset,
-	comHosts []dtclient.CommunicationHost,
-	role string, logger logr.Logger) bool {
+func (r *ReconcileOneAgent) reconcileIstioCreateConfigurations(instance *dynatracev1alpha1.OneAgent,
```
esukram: Just minor, but having the arguments separated by line, when so many arguments are given, increases readability.

pkg/controller/oneagent/istio.go
```diff
@@ -181,65 +191,93 @@ func (r *ReconcileOneAgent) removeIstioConfigurationForVirtualService(
...
+	istioClient *versionedistioclient.Clientset, communicationHosts []dtclient.CommunicationHost, role string,
+	logger logr.Logger) bool {
 
-	created := false
+	crdProbe := r.verifyIstioCrdAvailability(instance, logger)
+	if crdProbe != probeTypeFound {
+		logger.Info("istio: failed to lookup CRD for ServiceEntry/VirtualService: Did you install Istio recently? Please restart the Operator")
```
esukram: I would place a `.` or `!` at the end of the log.

pkg/controller/oneagent/istio.go
```diff
@@ -181,65 +191,93 @@ func (r *ReconcileOneAgent) removeIstioConfigurationForVirtualService(
...
+	if err != nil {
+		logger.Error(err, "istio: failed to create VirtualService")
+		return false
+	}
+	logger.Info("istio: VirtualService created", "objectName", name, "host", communicationHost.Host, "port", communicationHost.Port, "protocol", communicationHost.Protocol)
+
+	return true
+}
+
+func (r *ReconcileOneAgent) handleIstioConfigurationForServiceEntry(instance *dynatracev1alpha1.OneAgent, name string,
```
esukram: Similarly as above.

pkg/controller/oneagent/istio.go
```diff
@@ -181,65 +191,93 @@ func (r *ReconcileOneAgent) removeIstioConfigurationForVirtualService(
...
-	role string, logger logr.Logger) bool {
+func (r *ReconcileOneAgent) reconcileIstioCreateConfigurations(instance *dynatracev1alpha1.OneAgent,
+	istioClient *versionedistioclient.Clientset, communicationHosts []dtclient.CommunicationHost, role string,
+	logger logr.Logger) bool {
 
-	created := false
+	crdProbe := r.verifyIstioCrdAvailability(instance, logger)
+	if crdProbe != probeTypeFound {
+		logger.Info("istio: failed to lookup CRD for ServiceEntry/VirtualService: Did you install Istio recently? Please restart the Operator")
+		return false
```
esukram: Can't we _just_ exit the process and forcing a _restart_ of the operator process?

This of cause has a change of `CrashLoopBackOff` if this path in the code happens to be called more than just in that condition ;)

pkg/controller/oneagent/istio.go
namratachaudhary: since throughout the project the format for declaring multiple vars is the same, it would create some mismatch in the way things are defined in other files. 
We can probably rework this at a later stage, but for now - keeping it as it to maintain uniformity.

pkg/controller/oneagent/oneagent_controller.go
```diff
@@ -136,9 +138,9 @@ func (r *ReconcileOneAgent) Reconcile(request reconcile.Request) (reconcile.Resu
 	}
 	r.scheme.Default(instance)
 
-	if err := validate(instance); err != nil {
-		return reconcile.Result{}, err
-	}
+	// if err := validate(instance); err != nil {
```
lrgar: Is there a reason to comment this validation?

PR #151, opened 2019-10-01
Improve some error messages and comments

lrgar:


pkg/controller/istio/controller.go
```diff
@@ -285,7 +285,7 @@ func (c *Controller) createIstioConfigurationForServiceEntry(oneagent *dynatrace
 	serviceEntry.Labels = buildIstioLabels(oneagent.Name, role)
 	sve, err := c.istioClient.NetworkingV1alpha3().ServiceEntries(oneagent.Namespace).Create(serviceEntry)
 	if err != nil {
-		err = fmt.Errorf("istio: error listing service entries, %v", err)
+		err = fmt.Errorf("istio: error creating service entry, %v", err)
 		c.logger.Error(err, "istio reconcile")
```
baichinger: Returning the error from `Create` is sufficient. It gets logged by the caller anyways and logging it twice is pointless. Same for the other occurrences.

lrgar:
Will just pass the error along, since it gets logged by the caller.

As for the branch name, what is an appropriate one for these changes? I guess bugfix since it's correcting error messages/comments.

PR #152, opened 2019-10-02
Removing noise from logs

namratachaudhary:
As we watch the nodes, we get plenty of requests to the controller as state of the nodes keep changing. This causes the logs to be flooded. 

Since, all requests with namespace value empty are handled by nodes-controller, and we do not need to log every single request - we should only log the initialization of reconcile requests when received with adequate namespace - (onagent and istio reconcile workflow).


esukram:
Small but important change, it matter when to log ;)

PR #154, opened 2019-10-11
Backdate marked for termination event

baichinger:
A _marked for termination_ event sent after an agent stopped sending data is not considered by the problem detection algorithm. This happens in environments without [cluster-autoscaler](https://github.com/kubernetes/autoscaler/tree/master/cluster-autoscaler) due to nodes being shut down forcefully.

In order for this event to be picked up by the problem detection algorithm, the timestamp the event happened needs to be set to a value prior to the last message received from the agent. The proposed simple solution sets the timestamp of the event 10 minutes into the past.

### DIFF
--- a/pkg/controller/nodes/controller.go
+++ b/pkg/controller/nodes/controller.go
@@ -133,7 +133,7 @@ func (c *Controller) sendMarkedForTerminationEvent(dtc dtclient.Client, nodeIP s
 		return err
 	}
 
-	tenMinutesAgoInMillis := uint64(time.Now().Add(-10*time.Minute).UnixNano() / 1_000_000)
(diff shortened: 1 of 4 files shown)


pkg/controller/nodes/controller.go
```diff
@@ -132,11 +133,12 @@ func (c *Controller) sendMarkedForTerminationEvent(dtc dtclient.Client, nodeIP s
 		return err
 	}
 
+	tenMinutesAgoInMillis := uint64(time.Now().Add(-10*time.Minute).UnixNano() / 1_000_000)
```
lrgar: The CircleCI pipeline uses Go 1.12 and this is a Go 1.13's number literal which makes compilation fail. Added a commit to upgrade Go.

PR #156, opened 2019-10-23
Update installation instructions for OpenShift

lrgar:
Our `openshift.yaml` refers to the RHCC image in releases. To be able to pull images from RHCC you need to be authenticated against it. On OCP 4.x, the environment is already configured for this, but this may not be the case with OCP 3.x environments.

We do provide instructions on how to authenticate already for the OneAgent service account (used for the OneAgent pods) but we don't indicate the steps for the Operator service account itself. On this PR we're adding the commands to link the imagePullSecrets to the Operator service account.

However, the user adding those secrets after applying `openshift.yaml` would stop the Operator from running at the beginning until such secrets are added. So we'd like for the customer to create the secrets before deploying the Operator, but then the customer wouldn't be able to link the secrets to service accounts that haven't been created until `openshift.yaml` is applied.

To improve this, we're modifying the manifest so that the secrets are linked already, so the customer only needs to create the secrets. This step is optional if your environment is already prepared (e.g., OCP 4.)

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@ Depending of the version of the Dynatrace OneAgent Operator, it supports the fol
 | Dynatrace OneAgent Operator version | Kubernetes | OpenShift Container Platform |
 | ----------------------------------- | ---------- | ---------------------------- |
 | master                              | 1.11+      | 3.11+                        |
-| v0.4.0                              | 1.11+      | 3.11+                        |
+| v0.4.2                              | 1.11+      | 3.11+                        |
 | v0.3.1                              | 1.11-1.15  | 3.11+                        |
 | v0.2.1                              | 1.9-1.15   | 3.9+                         |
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -51,6 +51,21 @@ $ kubectl -n dynatrace logs -f deployment/dynatrace-oneagent-operator
 #### OpenShift
 ```sh
 $ oc adm new-project --node-selector="" dynatrace
+```
+
+If you are installing the Operator on an **OpenShift Container Platform 3.11** environment, in order to use the certified [OneAgent Operator](https://access.redhat.com/containers/#/registry.connect.redhat.com/dynatrace/dynatrace-oneagent-operator) and [OneAgent](https://access.redhat.com/containers/#/registry.connect.redhat.com/dynatrace/oneagent) images from [Red Hat Container Catalog](https://access.redhat.com/containers/) (RHCC), you need to [provide image pull secrets](https://access.redhat.com/documentation/en-us/openshift_container_platform/3.9/html/developer_guide/dev-guide-managing-images#pulling-private-registries-delegated-auth). The Service Accounts on the `openshift.yaml` manifest already have links to the secrets to be created below.
```
mreider: I think you could say say this more concisely, and I am not sure the links are necessary. I like short bits of text with fewer words.

https://www.evernote.com/l/AIAfY6BeAJtCMZzyVeCGM9mfGJ3ENZIiJZQ

PR #162, opened 2019-11-06
Allow custom DNS Policy for OneAgent pods

lrgar:
The created OneAgent pods use the default DNS policy, ClusterFirst. However, some customers may need to use a custom policy. The entry has been added as the field `dnsPolicy` on the OneAgent CR Spec.


pkg/controller/oneagent/util.go
```diff
@@ -89,6 +89,8 @@ func copyDaemonSetSpecToOneAgentSpec(dsSpec *appsv1.DaemonSetSpec, crSpec *dynat
 	}
 	// PriorityClassName
 	crSpec.PriorityClassName = dsSpec.Template.Spec.PriorityClassName
+	// DNSPolicy
```
namratachaudhary: _Guessing that this goes into master, so refactor changes can be incorporated._

What use are these comments? they are throughout the func. Redundant and can be removed.

pkg/integration_tests/oneagent_controller_integration_test.go
```diff
@@ -36,4 +37,33 @@ func TestReconcileOneAgent_ReconcileOnEmptyEnvironment(t *testing.T) {
 
 	assert.Equal(t, DefaultTestNamespace, dsActual.Namespace, "wrong namespace")
 	assert.Equal(t, oaName, dsActual.GetObjectMeta().GetName(), "wrong name")
+	assert.Equal(t, corev1.DNSClusterFirst, dsActual.Spec.Template.Spec.DNSPolicy) // default when empty.
```
namratachaudhary: nitpick - remove the comment here as well, please incorporate these into message args etc.

pkg/integration_tests/oneagent_controller_integration_test.go
```diff
@@ -36,4 +37,33 @@ func TestReconcileOneAgent_ReconcileOnEmptyEnvironment(t *testing.T) {
...
+	e.AddOneAgent(oaName, &dynatracev1alpha1.OneAgentSpec{
+		ApiUrl:    DefaultTestAPIURL,
+		Tokens:    "token-test",
+		DNSPolicy: corev1.DNSClusterFirstWithHostNet,
+	})
+
+	_, err = e.Reconciler.Reconcile(newReconciliationRequest(oaName))
+	assert.NoError(t, err, "error reconciling")
+
+	// Check if deamonset has been created and has correct namespace and name.
```
namratachaudhary: same as above.

pkg/integration_tests/oneagent_controller_integration_test.go
```diff
@@ -36,4 +37,33 @@ func TestReconcileOneAgent_ReconcileOnEmptyEnvironment(t *testing.T) {
 
 	assert.Equal(t, DefaultTestNamespace, dsActual.Namespace, "wrong namespace")
 	assert.Equal(t, oaName, dsActual.GetObjectMeta().GetName(), "wrong name")
+	assert.Equal(t, corev1.DNSClusterFirst, dsActual.Spec.Template.Spec.DNSPolicy) // default when empty.
+}
+
+func TestReconcileOneAgent_ReconcileOnEmptyEnvironmentAndDNSPolicy(t *testing.T) {
```
namratachaudhary: While this test is great, can we also look into incorporating unit tests once again, now that we discovered the fake clientset can be used from controller-runtime?

README.md
```diff
@@ -124,6 +124,8 @@ spec:
   # when enabled, and if Istio is installed on the Kubernetes environment, then the Operator will create the corresponding
   # VirtualService and ServiceEntries objects to allow access to the Dynatrace cluster from the agent.
   #enableIstio: false
+  # DNS Policy for OneAgent pods (optional.) Empty for default, other options: ClusterFirstWithHostNet, etc.
```
namratachaudhary: `other options: ClusterFirstWithHostNet, etc`

"etc" is not very helpful - either please link a documentation link or describe this with available options. I assume not everyone would memorize the available DNS Policies? :grin:

deploy/cr.yaml
```diff
@@ -49,3 +49,5 @@ spec:
   # when enabled, and if Istio is installed on the Kubernetes environment, then the Operator will create the corresponding
   # VirtualService and ServiceEntries objects to allow access to the Dynatrace cluster from the agent.
   #enableIstio: false
+  # DNS Policy for OneAgent pods (optional.) Empty for default, other options: ClusterFirstWithHostNet, etc.
```
namratachaudhary: Same as mentioned in Readme.

pkg/integration_tests/oneagent_controller_integration_test.go
lrgar: Assert messages are used to describe why something failed. I just added it as a reference since it's not obvious what is the default.

pkg/controller/oneagent/util.go
lrgar: From what I can see they are used for reference to recognize which attributes from the OneAgent spec are being taken care of when creating the DaemonSet. From my point of view, these are fine.

namratachaudhary: > From what I can see they are used for reference to recognize which attributes from the OneAgent spec are being taken care of when creating the DaemonSet

that should be done while defining the struct; not while setting fields for the same. i still suggest they seem redundant. 
For a special case when field and attribute names do not match, these comments can be added.

pkg/integration_tests/oneagent_controller_integration_test.go
namratachaudhary: I would, in that case, nest two tests - one with default behavior, and one without (assuming if you have added that comment, there is no test case demonstrating default behavior?)

PR #168, opened 2019-11-13
Use correct secret when creating client for Dynatrace API

lrgar:


pkg/controller/oneagent-utils/verification_test.go
```diff
@@ -62,3 +65,29 @@ func TestVerifySecret(t *testing.T) {
 		assert.NoError(t, err)
 	}
 }
+
+func TestBuildDynatraceClientWithCustomToken(t *testing.T) {
```
baichinger: Suggest to add an additional test for covering the erroneous use case.

lrgar: This is the erroneous use case, but I can add one for the normal one (i.e., token and secret have the same name.)

baichinger: This one asserts `NoError`. I would add one which asserts an `Error` to verify if a faulty configuration gets handled accordingly.

lrgar: Ah, ok, fair, will do.

PR #171, opened 2019-11-22
Update CRD according to openapiv3schema

DTMad:
Update CRD according to the openapiv3schema and adds descriptions to every parameter.

Furthermore defines conditions now in the status to have detailed information about the status of the custom resource.


pkg/controller/oneagent/oneagent_controller.go
```diff
@@ -309,18 +377,18 @@ func (r *ReconcileOneAgent) updateCR(instance *dynatracev1alpha1.OneAgent) error
 
 	newStatus := instance.Status
 
-	// Rather than sending the existing value, the .status.items map also gets replaced in-place, so we send a
-	// dummy object to avoid modifying it.
-	instance.Status = dynatracev1alpha1.OneAgentStatus{}
-
-	if err := r.client.Update(context.TODO(), instance); err != nil {
+	instance.Status = newStatus
```
lrgar: You don't really need this instruction. Or at least run it after `.Update()`. but don't think it would be used.

pkg/controller/oneagent/oneagent_controller.go
```diff
@@ -309,18 +377,18 @@ func (r *ReconcileOneAgent) updateCR(instance *dynatracev1alpha1.OneAgent) error
...
 
-	instance.Status = newStatus
-
-	// Now, with this call we do update the Status section to the new value.
-	return r.client.Status().Update(context.TODO(), instance)
+	// Rather than sending the existing value, the .status.items map also gets replaced in-place, so we send a
+	// dummy object to avoid modifying it.
+	instance.Status = dynatracev1alpha1.OneAgentStatus{}
+	r.logger.Info("Updating spec")
+	return r.client.Update(context.TODO(), instance)
```
lrgar: Just for confirmation,

`.Status().Update()` doesn't replace `.spec` to the old values, right? - `.Update()` does modify `.status`, which is why I ask.

pkg/apis/dynatrace/v1alpha1/oneagent_types.go
```diff
@@ -39,11 +47,43 @@ type OneAgentSpec struct {
...
+
+const (
+	ApiTokenConditionType  OneAgentConditionType = "ApiToken"
+	PaaSTokenConditionType OneAgentConditionType = "PaaSToken"
+)
+
+type OneAgentCondition struct {
+	Type   OneAgentConditionType  `json:"type"`
+	Status corev1.ConditionStatus `json:"status"`
+	// Unique, this should be a short, machine understandable string that gives the reason
```
lrgar: FYI, this unrelated comment is being included on the schema.

pkg/controller/oneagent/oneagent_controller.go
```diff
@@ -309,18 +377,18 @@ func (r *ReconcileOneAgent) updateCR(instance *dynatracev1alpha1.OneAgent) error
 
 	newStatus := instance.Status
 
-	// Rather than sending the existing value, the .status.items map also gets replaced in-place, so we send a
-	// dummy object to avoid modifying it.
-	instance.Status = dynatracev1alpha1.OneAgentStatus{}
-
-	if err := r.client.Update(context.TODO(), instance); err != nil {
+	instance.Status = newStatus
+	r.logger.Info("Updating status")
```
lrgar: Debug log entry.

pkg/controller/oneagent/oneagent_controller.go
```diff
@@ -309,18 +377,18 @@ func (r *ReconcileOneAgent) updateCR(instance *dynatracev1alpha1.OneAgent) error
...
 	}
 
-	instance.Status = newStatus
-
-	// Now, with this call we do update the Status section to the new value.
-	return r.client.Status().Update(context.TODO(), instance)
+	// Rather than sending the existing value, the .status.items map also gets replaced in-place, so we send a
+	// dummy object to avoid modifying it.
+	instance.Status = dynatracev1alpha1.OneAgentStatus{}
+	r.logger.Info("Updating spec")
```
lrgar: Debug log entry.

pkg/controller/oneagent/oneagent_controller.go
```diff
@@ -252,6 +295,31 @@ func (r *ReconcileOneAgent) reconcileRollout(logger logr.Logger, instance *dynat
...
+	err := r.client.Get(context.TODO(), types.NamespacedName{Name: instance.Name, Namespace: instance.Namespace}, dsActual)
+	if err != nil && !errors.IsNotFound(err) {
+		phaseChanged = instance.Status.Phase != dynatracev1alpha1.Error
+		instance.Status.Phase = dynatracev1alpha1.Error
+		return phaseChanged, err
+	}
+
+	// Case: Daemonset exists
+	if !errors.IsNotFound(err) {
+		if dsActual.Status.NumberReady == dsActual.Status.CurrentNumberScheduled {
```
lrgar: There are some other scenarios, such as, what if `CurrentNumberScheduled` equals 0 (i.e., can't deploy on any node) - but we can take them as we go on.

pkg/controller/oneagent/oneagent_controller.go
```diff
@@ -194,6 +198,45 @@ func (r *ReconcileOneAgent) Reconcile(request reconcile.Request) (reconcile.Resu
...
+func (r *ReconcileOneAgent) validateSecret(instance *dynatracev1alpha1.OneAgent) error {
+	secret := &corev1.Secret{}
+	err := r.client.Get(context.TODO(), client.ObjectKey{Namespace: instance.Namespace, Name: instance.Spec.Tokens}, secret)
+	if err != nil && !errors.IsNotFound(err) {
+		return err
+	}
+
+	if errors.IsNotFound(err) {
+		var messageApiToken = fmt.Sprintf("The secret %s has not been found on namespace %s", instance.Spec.Tokens, instance.Namespace)
+		var messagePaasToken = fmt.Sprintf("The secret %s has not been found on namespace %s", instance.Spec.Tokens, instance.Namespace)
```
lrgar: It's the same message if you want to avoid duplication.

pkg/controller/oneagent/oneagent_controller.go
```diff
@@ -252,6 +295,31 @@ func (r *ReconcileOneAgent) reconcileRollout(logger logr.Logger, instance *dynat
 	return updateCR, nil
 }
 
+func (r *ReconcileOneAgent) determineOneAgentPhase(instance *dynatracev1alpha1.OneAgent) (bool, error) {
+	// Check if this DaemonSet already exists
```
namratachaudhary: nitpick : unnecessary comment on L#299 and L#309

pkg/controller/oneagent/oneagent_controller.go
```diff
@@ -252,6 +295,31 @@ func (r *ReconcileOneAgent) reconcileRollout(logger logr.Logger, instance *dynat
...
+	dsActual := &appsv1.DaemonSet{}
+	err := r.client.Get(context.TODO(), types.NamespacedName{Name: instance.Name, Namespace: instance.Namespace}, dsActual)
+	if err != nil && !errors.IsNotFound(err) {
+		phaseChanged = instance.Status.Phase != dynatracev1alpha1.Error
+		instance.Status.Phase = dynatracev1alpha1.Error
+		return phaseChanged, err
+	}
+
+	// Case: Daemonset exists
+	if !errors.IsNotFound(err) {
```
namratachaudhary: this is already being checked for, at 303.. 

```
if err != nil && !errors.IsNotFound(err)
```
and
```
if !errors.IsNotFound(err) {}
```
are basically same conditions.

lrgar: Hmm, they are different: the first one says "any error that is not NotFound" and the second "the error is not NotFound" (i.e., exists), the leftover case is that the error is NotFound (i.e., doesn't exist)

pkg/apis/dynatrace/v1alpha1/oneagent_types.go
DTMad: Removed unnecessary comment

pkg/controller/oneagent/oneagent_controller.go
DTMad: Remove unnecessary debug log line

pkg/controller/oneagent/oneagent_controller.go
DTMad: Remove unnecessary debug log line

pkg/controller/oneagent/oneagent_controller.go
DTMad: Removed message once and reused the other

pkg/controller/oneagent/oneagent_controller.go
DTMad: Removed comments

pkg/controller/oneagent/oneagent_controller.go
namratachaudhary: // Check if this DaemonSet already exists
	// Case: Daemonset exists
￼How are they different :confused: 
I would say split it logically if they are really that different, but imo, it does not seem so.

PR #174, opened 2019-11-27
Use istio.io/client-go library to manage Istio objects

lrgar:
When first implemented, there was no official library we could use to create and manage the Istio objects for Kubernetes, so we generated our own and included it in our repository.

There is one now, https://github.com/istio/client-go, so this PR replaces the internal usage to this library.


lrgar:
Rebased. Please re-approve.

PR #179, opened 2019-12-03
Add Dockerfile based on UBI base image

lrgar:
This PR uses an UBI-based Dockerfile for the Operator for both Quay and RHCC images.

The Dockerfile and the new helper files are from the template created by the operator-sdk v0.12 CLI, but  I modified the `ubi-minimal:latest`'s tag to `8.1` instead.

Note that there is now an `ENTRYPOINT` instruction on the Dockerfile, so no need to specify the command on the deployment manifests. (Although, curiously, the command field is still mentioned on the generated deployments from the operator-sdk)


lrgar:
Rebased from master.

PR #181, opened 2019-12-05
upload manifests as attachment to release

thomschuetz:


.circleci/config.yml
```diff
@@ -134,6 +134,24 @@ jobs:
...
+          current_image="dynatrace-agent-operator:${CIRCLE_TAG}"
+          mkdir artefacts
+          sed "s/${template_image}/${current_image}/g" deploy/kubernetes.yaml > artefacts/kubernetes.yaml
+          sed "s/${template_image}/${current_image}/g" deploy/kubernetes.yaml > artefacts/openshift.yaml
+      - attach_workspace:
+          at: .
+      - run: |
+          echo $CIRCLE_TAG
+          ls artefacts/
+          # ghr -t ${GITHUB_TOKEN} -u ${CIRCLE_PROJECT_USERNAME} -r ${CIRCLE_PROJECT_REPONAME} -c ${CIRCLE_SHA1} -delete ${CIRCLE_TAG} artefacts/
```
lrgar: Unused comment?

.circleci/config.yml
```diff
@@ -134,6 +134,24 @@ jobs:
...
+    docker:
+      - image: cibuilds/github
+    steps:
+      - checkout
+      - run: |
+          template_image="dynatrace-oneagent-operator:snapshot"
+          current_image="dynatrace-agent-operator:${CIRCLE_TAG}"
+          mkdir artefacts
+          sed "s/${template_image}/${current_image}/g" deploy/kubernetes.yaml > artefacts/kubernetes.yaml
+          sed "s/${template_image}/${current_image}/g" deploy/kubernetes.yaml > artefacts/openshift.yaml
```
lrgar: Use openshift.yaml as base file, plus the one on master uses Quay whereas the releases are on RHCC, so the image registry needs to be replaced as well.

.circleci/config.yml
thomschuetz: fixed

.circleci/config.yml
thomschuetz: fixed

PR #183, opened 2019-12-18
Labels for OneAgent CR

namratachaudhary:


pkg/apis/dynatrace/v1alpha1/oneagent_types.go
```diff
@@ -45,6 +45,8 @@ type OneAgentSpec struct {
 	EnableIstio bool `json:"enableIstio,omitempty"`
 	// DNS Policy for the OneAgent pods.
 	DNSPolicy corev1.DNSPolicy `json:"dnsPolicy,omitempty"`
+	// Labels for the OneAgent pods
+	Labels map[string]string `json:"labels, omitempty"`
```
lrgar: Hmm, I think the whitespace before omitempty marking this field as required on the CRD above.

pkg/controller/oneagent/oneagent_controller.go
```diff
@@ -384,13 +384,18 @@ func (r *ReconcileOneAgent) updateCR(instance *dynatracev1alpha1.OneAgent) error
...
+		}
+	}
+
 	podSpec := newPodSpecForCR(instance)
 
 	return &appsv1.DaemonSet{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      instance.Name,
 			Namespace: instance.Namespace,
-			Labels:    selector,
```
lrgar: Do we want to not have any labels for the DS?

pkg/controller/oneagent/oneagent_utils.go
```diff
@@ -85,6 +86,7 @@ func copyDaemonSetSpecToOneAgentSpec(dsSpec *appsv1.DaemonSetSpec, crSpec *dynat
 	}
 	crSpec.PriorityClassName = dsSpec.Template.Spec.PriorityClassName
 	crSpec.DNSPolicy = dsSpec.Template.Spec.DNSPolicy
+	crSpec.Labels = dsSpec.Template.Labels
```
lrgar: I've the strong impression that this will force the Operator to update the DS on every reconciliation, because the DS labels won't match what it's on CR. Look at `hasSpecChanged()` for implementation details.

I suggest leaving this line here, but moving the labels generation from `newDaemonSetForCR()` to `SetDefaults_OneAgentSpec()` in `default.go`.

pkg/controller/oneagent/oneagent_controller.go
namratachaudhary: This is not allowed, somehow. Throws an error ; refer - https://circleci.com/gh/Dynatrace/dynatrace-oneagent-operator/1070?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

lrgar:
```
pkg/controller/oneagent/oneagent_controller.go:400: cannot use promoted field ObjectMeta.Labels in struct literal of type "k8s.io/api/core/v1".PodTemplateSpec
```

This was for this line: https://github.com/Dynatrace/dynatrace-oneagent-operator/pull/183/commits/4bb823971fab36e3e1355da6fa40bde23c19457d#diff-a0ddcc1928501dbec12491bace283fa3R400 which was not the 393 mentioned above.

pkg/controller/oneagent/oneagent_utils.go
lrgar: Since `SetDefaults_OneAgentSpec()` doesn't seem to work since the name is not available, then I suggest `ReconcileOneAgent.Reconcile()` instead, as we are doing with the tokens already.

pkg/controller/oneagent/oneagent_controller.go
```diff
@@ -131,6 +131,8 @@ func (r *ReconcileOneAgent) Reconcile(request reconcile.Request) (reconcile.Resu
 	}
 	r.scheme.Default(instance)
 
+	instance.Labels = mergeLabels(instance.GetLabels(), buildLabels(instance.Name))
```
lrgar:
```suggestion
	instance.Spec.Labels = mergeLabels(instance.Spec.Labels, buildLabels(instance.Name))
```

`instance.Labels` and `instance.GetLabels()` both return `.metadata.labels` and not `.spec.labels` AFAIK.

The same applies to `newDaemonSetForCR()` but that aside it should do.

Although perhaps we need to handle `instance.Spec.Labels == nil`

PR #185, opened 2019-12-27
Improve error logging from Dynatrace API requests on Operator

namratachaudhary:


pkg/controller/istio/controller.go
```diff
@@ -65,61 +65,67 @@ func (c *Controller) initialiseIstioClient(config *rest.Config) (istioclientset.
...
 }
 
 func (c *Controller) reconcileIstioConfigurations(instance *dynatracev1alpha1.OneAgent,
 	comHosts []dtclient.CommunicationHost, role string) (bool, error) {
 
-	add := c.reconcileIstioCreateConfigurations(instance, comHosts, role)
-	rem := c.reconcileIstioRemoveConfigurations(instance, comHosts, role)
+	add, err := c.reconcileIstioCreateConfigurations(instance, comHosts, role)
+	if err != nil {
+		return false, nil
```
lrgar:
```suggestion
		return false, err
```

pkg/controller/istio/controller.go
```diff
@@ -65,61 +65,67 @@ func (c *Controller) initialiseIstioClient(config *rest.Config) (istioclientset.
...
+	}
+	rem, err := c.reconcileIstioRemoveConfigurations(instance, comHosts, role)
+	if err != nil {
+		return false, nil
```
lrgar:
```suggestion
		return false, err
```

pkg/controller/oneagent/oneagent_controller.go
```diff
@@ -290,7 +294,7 @@ func (r *ReconcileOneAgent) reconcileRollout(logger logr.Logger, instance *dynat
 		desired, err := dtc.GetLatestAgentVersion(dtclient.OsUnix, dtclient.InstallerTypeDefault)
 		if err != nil {
 			logger.Error(err, "failed to get desired version")
-			return updateCR, nil
+			return updateCR, err
```
lrgar: Just logging and not returning an error was on purpose here I think, since we can do some work even if we can't access the API.

pkg/controller/oneagent/oneagent_controller.go
```diff
@@ -333,7 +337,7 @@ func (r *ReconcileOneAgent) reconcileVersion(logger logr.Logger, instance *dynat
 	desired, err := dtc.GetLatestAgentVersion(dtclient.OsUnix, dtclient.InstallerTypeDefault)
 	if err != nil {
 		logger.Error(err, "failed to get desired version")
-		return false, nil
+		return false, err
```
lrgar: Same here.

pkg/controller/oneagent/oneagent_controller.go
```diff
@@ -164,7 +164,11 @@ func (r *ReconcileOneAgent) Reconcile(request reconcile.Request) (reconcile.Resu
 	}
 
 	if instance.Spec.EnableIstio {
-		if upd, ok := r.istioController.ReconcileIstio(instance, dtc); ok && upd {
+		upd, ok, err := r.istioController.ReconcileIstio(instance, dtc)
+		if !ok && err != nil {
+			return reconcile.Result{}, err
```
lrgar: Just log the error, but don't return. If Istio configuration fails, it doesn't necessarily break the rest of the OneAgent.

pkg/dtclient/dynatrace_client.go
```diff
@@ -64,25 +64,35 @@ func (dc *dynatraceClient) makeRequest(url string, tokenType tokenType) (*http.R
 	return dc.httpClient.Do(req)
 }
 
-func (dc *dynatraceClient) getServerResponseData(response *http.Response) ([]byte, error) {
+func (dc *dynatraceClient) getServerResponseData(response *http.Response) ([]byte, *serverError, error) {
```
lrgar: You can use Go 1.13 errors to manage specific types of errors. I'm currently working on some similar changes for this, so we can leave it as it's now.

lrgar:
Please add an entry to CHANGELOG.md.

PR #194, opened 2020-01-20
Use reader from mgr.GetApiReader() for querying the OneAgent object

mmayr-at:
Querying the OneAgent object was done with the client received by `mgr.GetClient()`. For fast reconciliation cycles, where the OneAgent object is modified and read again, the client might return an out of fdate OneAgent object from the cahce, resulting in following error message:
```json
{
    "level": "error",
    "ts": 1573211855.705202,
    "logger": "kubebuilder.controller",
    "msg": "Reconciler error",
    "controller": "oneagent-controller",
    "request": "dynatrace/oneagent",
    "error": "Operation cannot be fulfilled on oneagents.dynatrace.com \"oneagent\": the object has been modified; please apply your changes to the latest version and try again",
}
```

To prevent this issue a client instance that does not use caching is provided via `mgr.GetApiReader()`. This instance is now used to query the OneAgent object in the reconciliation loop.

### DIFF
--- a/pkg/controller/oneagent/oneagent_controller.go
+++ b/pkg/controller/oneagent/oneagent_controller.go
@@ -45,16 +45,17 @@ func Add(mgr manager.Manager) error {
 
 // newReconciler returns a new reconcile.Reconciler
 func newReconciler(mgr manager.Manager) reconcile.Reconciler {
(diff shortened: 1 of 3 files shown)


lrgar:
Restarted the build since it failed to fetch the base image from RHCC registry, temporary thing I presume.

Feel free to merge.

PR #197, opened 2020-01-28
Set status phase value during the reconciliation loop

mmayr-at:
This PR adds a value for `status.phase` to the OneAgent object. The value is set to 
* `Error` in case any error happens during the reconciliation loop
* `Deploying` when the operator updates the OneAgent pods/DaemonSet
* `Running` when everything is up to date and running

### DIFF
--- a/pkg/controller/oneagent/oneagent_controller_test.go
+++ b/pkg/controller/oneagent/oneagent_controller_test.go
@@ -2,6 +2,7 @@ package oneagent
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"os"
 	"testing"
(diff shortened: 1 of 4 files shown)


pkg/controller/oneagent/oneagent_controller_test.go
```diff
@@ -252,6 +253,96 @@ func TestReconcileDynatraceClient_ProbeRequests(t *testing.T) {
 	})
 }
 
+func TestReconcile_PhaseSetCorrectly(t *testing.T) {
+	//now := metav1.Now()
```
DTMad: Unused comment can be removed ;)

lrgar:
Oh, and please add an entry to `CHANGELOG.md` before merging.

PR #201, opened 2020-02-07
Add build config for TravisCI

DTMad:


.travis.yml
```diff
@@ -0,0 +1,215 @@
...
+          - $HOME/.cache/go-build
+          - $HOME/gopath/pkg/mod
+
+  - &deploy
+    script: ./build/deploy.sh
+
+  - &create_manifest
+    stage: manifest
+    install: skip
+    git:
```
lrgar: I guess this needs `language: minimal` or `generic` as well sinc Git clone is disabled: https://docs.travis-ci.com/user/languages/minimal-and-generic/

.travis.yml
```diff
@@ -0,0 +1,215 @@
...
+      env: IMAGE=$OAO_IMAGE_DOCKER_HUB TAG="snapshot"
+      workspaces:
+        use:
+          - ws_amd
+          - dockerconfig
+
+    - <<: *deploy
+      name: Deploy snapshot quay AMD
+      arch: amd64
+      if: branch != master
```
lrgar: I wonder if `branch != master` also includes release tags?

.travis.yml
DTMad: Changed to generic

.travis.yml
DTMad: Added an additional check if the tag is blank

PR #216, opened 2020-03-16
Add handling for reaching the Dynatrace API quotas

DTMad:
Due to our limitation for a max of 50 requests per minute to our Dynatrace API our Operator sometimes displayed error messages.
Implemented proper error handling and added a reconciliation interval of 1 minute in this case.

### DIFF
--- a/pkg/controller/oneagent/oneagent_controller.go
+++ b/pkg/controller/oneagent/oneagent_controller.go
@@ -181,6 +181,12 @@ func (r *ReconcileOneAgent) Reconcile(request reconcile.Request) (reconcile.Resu
 		}
 
 		return reconcile.Result{RequeueAfter: 5 * time.Minute}, nil
+	} else if err != nil {
+		var serr dtclient.ServerError
+		if ok := errors.As(err, &serr); ok && serr.Code == http.StatusTooManyRequests {
+			logger.Info("Request limit for Dynatrace API reached! Next reconcile in one minute")
+			return reconcile.Result{RequeueAfter: 1 * time.Minute}, nil
+		}
 	}
 
 	if instance.Spec.DisableAgentUpdate {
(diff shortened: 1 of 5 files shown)


pkg/controller/oneagent/oneagent_controller.go
```diff
@@ -191,10 +202,21 @@ func (r *ReconcileOneAgent) Reconcile(request reconcile.Request) (reconcile.Resu
...
 			return reconcile.Result{}, err
 		}
 
 		return reconcile.Result{RequeueAfter: 5 * time.Minute}, nil
+	} else if err != nil {
+		var serr dtclient.ServerError
+		if ok := errors.As(err, &serr); ok && serr.Code == http.StatusTooManyRequests {
+			logger.Info("Request limit for Dynatrace API reached! Next reconcile in one minute")
+			return reconcile.Result{RequeueAfter: 1 * time.Minute}, nil
+		}
```
lrgar:
```suggestion
		}
		return reconcile.Result{}, err
```

pkg/controller/oneagent/oneagent_controller.go
```diff
@@ -172,10 +172,21 @@ func (r *ReconcileOneAgent) Reconcile(request reconcile.Request) (reconcile.Resu
...
```
lrgar:
```suggestion
		}
		reconcile.Result{}, err
```

pkg/controller/oneagent/oneagent_controller.go
```diff
@@ -245,8 +267,7 @@ func (r *ReconcileOneAgent) reconcileRollout(logger logr.Logger, instance *dynat
 	if instance.Status.Version == "" {
 		desired, err := dtc.GetLatestAgentVersion(dtclient.OsUnix, dtclient.InstallerTypeDefault)
 		if err != nil {
-			logger.Error(err, "failed to get desired version")
-			return updateCR, nil
+			return updateCR, fmt.Errorf("failed to get desired version: %w", err)
```
lrgar: This is technically a change of semantics, reconcileVersion() wouldn't now run if failed to get the latest version from the API. I think it's fine however.

DTMad: Should we adapt it?

lrgar: Let's leave it for a follow-up refactor. Although to be fair this is probably better anyway.

lrgar:
Rebase from master please.

pkg/controller/oneagent/oneagent_controller.go
DTMad: Will leave it then :)

PR #227, opened 2020-04-02
Prepare Operator for release to RedHat marketplace

DTMad:
To be prepared for the RedHat marketplace we added the image for the OneAgent as an environment variable in the CSV file for OpenShift and check in the code if it's set and use it.

### DIFF
--- a/deploy/olm/openshift/0.7.0/dynatrace-monitoring.v0.7.0.clusterserviceversion.yaml
+++ b/deploy/olm/openshift/0.7.0/dynatrace-monitoring.v0.7.0.clusterserviceversion.yaml
@@ -1,0 +1,437 @@
+apiVersion: operators.coreos.com/v1alpha1
+kind: ClusterServiceVersion
+metadata:
+  annotations:
+    alm-examples: |-
+      [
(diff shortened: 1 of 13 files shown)


deploy/olm/openshift/0.7.0/dynatrace-monitoring.v0.7.0.clusterserviceversion.yaml
```diff
@@ -0,0 +1,439 @@
...
+              - env:
+                - name: WATCH_NAMESPACE
+                  valueFrom:
+                    fieldRef:
+                      fieldPath: metadata.annotations['olm.targetNamespaces']
+                - name: POD_NAME
+                  valueFrom:
+                    fieldRef:
+                      fieldPath: metadata.name
+                - name: RELATED_IMAGE_DYNATRACE_ONEAGENT
```
lrgar, line 308: Remove it from here since 0.7.0 was released and doesn't have this. Just add it to the openshift/deployment.yaml deployment. It should be included in the v0.7.1 CSV once it gets generated.

PR #228, opened 2020-04-03
Prepare for RedHat marketplace release

DTMad:


lrgar:
Looks good. Please update the changelog with a new future section.

I guess we also need to update the CR examples in the Readme, cr.yaml, the CRD Yaml, and oneagent_types.go since they mention the DockerHub image as the default which would no longer be true in OpenShift.

PR #230, opened 2020-04-07
Fix Travis CI pipeline issues

DTMad:


.travis.yml
```diff
@@ -124,35 +107,37 @@ jobs:
           - ws_amd
           - dockerconfig
 
-    - <<: *deploy
-      name: Deploy snapshot quay AMD
-      arch: amd64
-      if: (branch != master) AND (tag IS blank)
+    - <<: *deploy_amd
```
lrgar, line 110: We're no longer publishing to Quay for releases, only development builds.

lrgar:
Please also add an entry to changelog.

.travis.yml
DTMad: Changed condition to "if: (branch != master) AND (tag IS blank)"

lrgar: Suggest renaming it to "Deploy snapshot quay AMD"

Please also do the name/if changes to the ARM job.

DTMad: Thanks for the hint! Done

PR #238, opened 2020-04-24
Use kustomize to create deployment manifests

DTMad:
To reduce maintenance overhead the kubernetes.yaml and  openshift.yaml got removed from the master branch and will just be created with kustomize for releases.


deploy/test.yaml
```diff
@@ -1,228 +1,3 @@
-apiVersion: v1
```
lrgar, line 1: Guess this file was for testing and can be removed.

DTMad: Removed

DTMad:
Changelog entry added + test files removed

PR #250, opened 2020-05-11
Work around occasionally missing webhook reconciliations

lrgar:
There is an issue on https://github.com/kubernetes-sigs/controller-runtime/issues/942, where pre-existing elements on the `source.Channel`'s channel used by the webhook's bootstrapper may not be processed.

As a workaround, I'm here delaying a bit before we generate the initial request, and artificially generating periodic reconciliations rather than leaving it to the ReconcileAfter (since, if the first request is not processed, then there would be no follow-ups.)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 ### Future
 
 #### Features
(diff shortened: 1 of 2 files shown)


scuzhanglei:
@lrgar  hi, is it lose events not only in webhook, and recocilers too??

PR #251, opened 2020-05-11
Rename OneAgent app-only directory

lrgar:
If we use `/opt/dynatrace/oneagent` to store the PaaS agent, and the full-stack agent is installed as well, then the init container will try to overwrite the files from the full-stack agent with the PaaS files, which will fail because of the lack of permissions (or not if the init container is running as root.)

While we don't define an expected behavior for this scenario, better to use different directories to avoid breaking stuff.

Using `/opt/dynatrace/oneagent-paas` for now.


esukram:
Luis, why not having the paas directory configurable via CR and defaulting to your suggestion?

Markus

lrgar:
> Luis, why not having the paas directory configurable via CR and defaulting to your suggestion?

May I ask about the use case? For example, if it would be a workaround for some conflicts with certain containers, then that sounds more like a Pod-specific thing and would better be configurable as a Pod annotation than to apply to all Pods.

lrgar:
Followed-up with Marco and decided to support an annotation on the Pods to allow configuration to a different path, e.g., `oneagent.dynatrace.com/install-path: "/custom/dir"` so will modify this PR accordingly.

lrgar:
PR updated, the customer containers will now have the agent on the `/opt/dynatrace/oneagent-paas` directory by default, but can be configured with the `oneagent.dynatrace.com/install-path` annotation on the Pods.

The init container however has been modified to only use `/mnt/oneagent` to extract the files to, since the directory shouldn't matter much for this container.

PR #252, opened 2020-05-12
Use a single binary to run the different Operator processes

lrgar:
We currently run three containers with the Operator package,
* The good old Operator itself.
* The webhook bootstrapper.
* The webhook server.

I initially created one binary for each one of these, but we can merge them into a single binary, and use a subcommand to run it on any of these modes. The difference between them is which reconcilers are started.

I'm calling the subcommands respectively,
* `operator`
* `webhook-bootstrapper`
* `webhook-server`

So, if you run the Operator locally, it would be like,

```
$ operator-sdk run --local --operator-flags="operator"
$ WATCH_NAMESPACE=dynatrace go run ./cmd/manager/ -- operator
```

Finally, I'm removing the `add_***.go` files that are created by the operator-sdk's scaffolding, and instead using a list on `cmd/manager/operator.go`. The idea here is that we don't really use scaffolding and just copy/paste for new reconcilers, and having a simple list is more clear than depending on the behavior of `init()` functions. The webhook processes are already just adding specific reconcilers, so we would get some consistency.


lrgar:
Now using `operator` as the default subcommand if nothing is set. You need to explicitly use the subcommand to run the webhook server and bootstrapper.

PR #254, opened 2020-05-14
Simplify Kustomize patches

lrgar:
By taking advantage of Kustomize's patchesJson6902, we can remove some duplicated YAMLs and just applying the changes we need.


lrgar:
Updated to use patch files instead of inline patches, since it doesn't seem to be supported by `kubectl apply -k` yet.

PR #270, opened 2020-06-04
Add NetworkZone and querying by network zone

DTMad:


pkg/dtclient/dynatrace_client.go
```diff
@@ -146,15 +147,18 @@ func (dc *dynatraceClient) setHostCacheFromResponse(response []byte) error {
...
-		if v == nil {
-			continue
-		}
-
-		hostInfo.version = fmt.Sprintf("%d.%d.%d.%s", v.Major, v.Minor, v.Revision, v.Timestamp)
-		for _, ip := range info.IPAddresses {
-			dc.hostCache[ip] = hostInfo
+		nz := info.NetworkZoneID
+
+		if nz == networkZone || nz == "default" {
```
lrgar, line 152: If you set the network zone, we should ignore hosts that don't have a network set.

```suggestion
		if nz == networkZone || networkZone == "" {
```

pkg/dtclient/client.go
```diff
@@ -33,7 +33,7 @@ type Client interface {
 	//
 	// The list of all hosts with their IP addresses is cached the first time this method is called. Use a new
 	// client instance to fetch a new list from the server.
-	GetAgentVersionForIP(ip string) (string, error)
+	GetAgentVersionForIP(ip string, networkZone string) (string, error)
```
lrgar, line 36: Alternatively, you can pass it on the Dynatrace client constructor as we already do for proxy, certs, etc., if you don't want to forward the argument everywhere, but as you prefer.

deploy/cr.yaml
```diff
@@ -66,3 +66,7 @@ spec:
   # Provide the name of the configmap which holds your .pem in a field called 'certs'
   # If this is not set the default embedded certificates on the images will be used
   #trustedCAs: name-of-my-ca-configmap
+  # Sets a NetworkZone for the OneAgent (optional)
+  # If this is not set the default NetworkZone is 'default'
```
lrgar, line 70: I'd say that the default is a technical detail and we shouldn't mention it here. It doesn't matter to us either, since we just don't set the installer argument in this case.

PR #279, opened 2020-07-23
Implemented "mark for termination" for cordoned off nodes

meik99:


lrgar:
A few things:
* We need to recognize also on Node changes, the reconcileAll happens every five minutes, which is not fast enough for the problem we want to solve.
* Once it has been sent, we don't want to keep sending the messages continuously, so we'll likely need to update the cache accordingly to indicate that a message has been sent - even perhaps avoiding sending the message on deletion if we sent the message recent enough.

pkg/controller/nodes/nodes_controller.go
```diff
@@ -288,3 +358,66 @@ func (r *ReconcileNodes) sendMarkedForTermination(oa *dynatracev1alpha1.OneAgent
...
+	oneAgent, err := r.determineOneAgentForNode(node.Name)
+	if err != nil {
+		return err
+	}
+	if oneAgent == nil {
+		return nil
+	}
+
+	instance, hasNodeInstance := oneAgent.Status.Instances[node.Name]
+	if !hasNodeInstance {
```
lrgar, line 372: As determineOneAgentForNode() currently implemented, i.e., check for this Instances list and see if it's listed - this won't happen.

pkg/controller/nodes/watchers.go
```diff
@@ -33,6 +33,43 @@ func (r *ReconcileNodes) watchDeletions(stop <-chan struct{}) (chan string, erro
...
+	chUpdates := make(chan map[string]string, 1)
+
+	informer.AddEventHandler(toolscache.ResourceEventHandlerFuncs{
+		UpdateFunc: r.handleUpdate(chUpdates),
+	})
+
+	return chUpdates, nil
+}
+
+func (r *ReconcileNodes) handleUpdate(chUpdates chan map[string]string) func(oldObj, newObj interface{}) {
```
lrgar, line 51: You can use a channel of strings since, again, names can't change.

.gitignore
```diff
@@ -79,3 +79,4 @@ tags
 # End of https://www.gitignore.io/api/go,vim,emacs,visualstudiocode
 .idea/
 vendor/
+/log.txt
```
DTMad, line 82: Added on purpose?

pkg/controller/nodes/nodes_controller.go
```diff
@@ -79,6 +88,10 @@ func (r *ReconcileNodes) Start(stop <-chan struct{}) error {
 			if err := r.onDeletion(node); err != nil {
 				r.logger.Error(err, "failed to reconcile deletion", "node", node)
 			}
+		case nodeMap := <-chUpdates:
```
lrgar, line 91: No longer a nodeMap.

pkg/controller/nodes/oneagent_dao.go
```diff
@@ -0,0 +1,53 @@
+package nodes
```
DTMad, line 1: oneagent_dao.go - what's dao?

pkg/controller/nodes/nodes_controller.go
```diff
@@ -87,17 +100,39 @@ func (r *ReconcileNodes) Start(stop <-chan struct{}) error {
...
+	logger := r.logger.WithValues("node", nodeName)
+	logger.Info("node update notification received")
+
+	c, err := r.getCache()
+	if err != nil {
+		return err
+	}
+
+	err = r.updateNode(c, nodeName)
+	if err != nil {
```
lrgar, line 113: When you aren't defining new variables, I suggest putting the operation inside the if.

It depends on the case, but if it's just for error validation, it saves vertical space.

```suggestion
	if err = r.updateNode(c, nodeName); err != nil {
```

pkg/controller/nodes/nodes_controller.go
```diff
@@ -87,17 +100,39 @@ func (r *ReconcileNodes) Start(stop <-chan struct{}) error {
...
+		return err
+	}
+
+	err = r.updateCache(c)
+	if err != nil {
+		return err
+	}
+
+	return nil
```
lrgar, line 122: You can return the error directly.

```suggestion
	return r.updateCache(c)
```

pkg/controller/nodes/nodes_controller.go
```diff
@@ -265,6 +311,23 @@ func (r *ReconcileNodes) removeNode(c *Cache, node string, oaFunc func(name stri
 	return nil
 }
 
+func (r *ReconcileNodes) updateNode(cache *Cache, nodeName string) error {
```
lrgar, line 314: Perhaps you want to use `c` for the `cache` variable as well, as it's now used on the rest of the PR?

pkg/controller/nodes/nodes_controller.go
```diff
@@ -288,3 +351,85 @@ func (r *ReconcileNodes) sendMarkedForTermination(oa *dynatracev1alpha1.OneAgent
...
+	return lastMarked.UTC().Add(time.Hour).Before(time.Now().UTC())
+}
+
+func updateLastMarkedForTerminationTimestamp(c *Cache, nodeInfo *CacheEntry, nodeName string) error {
+	nodeInfo.LastMarkedForTermination = time.Now().UTC()
+	err := c.Set(nodeName, *nodeInfo)
+	if err != nil {
+		return err
+	}
+	return nil
```
lrgar, line 434:
```suggestion
	return c.Set(nodeName, *nodeInfo)
```

pkg/controller/nodes/nodes_controller.go
```diff
@@ -288,3 +351,85 @@ func (r *ReconcileNodes) sendMarkedForTermination(oa *dynatracev1alpha1.OneAgent
...
+	}
+	if oneAgent == nil {
+		return nil
+	}
+
+	// determineOneAgentForNode  only returns a oneagent object if a node instance is present
+	instance, _ := oneAgent.Status.Instances[node.Name]
+	cachedNode, err := cache.Get(node.Name)
+	if err != nil {
+		if err == ErrNotFound {
```
lrgar, line 368: I'm curious here. Is there any scenario where an Update notification would be processed after a Deletion? It would re-add the entry and cause for the message to be sent twice.

I guess if a Update notification came while reconcileAll was running, but before listing all nodes.

Shouldn't break anything however. I know it was mentioned that we currently see the messages sent twice occasionally, but the timing needed here shouldn't happen that often even.

On the other hand, if I'm not mistaken, the case handled here would be when we get an update for an entry before it was caught up on reconcileAll() but for that to happen the Node needs to have been lived for less than 5 minutes before being deleted.

Don't think we need to concern ourselves with this scenario. Or perhaps is there another case handled here?

pkg/controller/nodes/nodes_controller.go
```diff
@@ -288,3 +351,85 @@ func (r *ReconcileNodes) sendMarkedForTermination(oa *dynatracev1alpha1.OneAgent
...
+
+	if isMarkableForTermination(&cachedNode) {
+		err = updateLastMarkedForTerminationTimestamp(c, &cachedNode, nodeName)
+		if err != nil {
+			return err
+		}
+
+		return r.sendMarkedForTermination(oneAgent, ipAddress, cachedNode.LastSeen)
+	}
+	return nil
```
lrgar, line 402: It's a Go pattern to exit quickly.

```suggestion
	if !isMarkableForTermination(&cachedNode) {
		return nil
	}

	if err = updateLastMarkedForTerminationTimestamp(c, &cachedNode, nodeName); err != nil {
		return err
	}

	return r.sendMarkedForTermination(oneAgent, ipAddress, cachedNode.LastSeen)
```

pkg/controller/nodes/nodes_controller_test.go
```diff
@@ -179,3 +105,82 @@ func TestNodesReconciler_NodeNotFound(t *testing.T) {
 	_, err := nodesCache.Get("node2")
 	assert.Equal(t, err, ErrNotFound)
 }
+
+func TestNodeReconciler_NodeHasTaint(t *testing.T) {
+	fakeClient := createDefaultFakeClientWithScheme()
+	dtc := createDTMockClient("1.2.3.4", "HOST-42")
```
lrgar, line 111: `dtc` is used here, while `dtClient` is used on other functions, if you want to be consistent. Same with `ctrl` and `reconciler`.

pkg/controller/nodes/watchers.go
```diff
@@ -33,6 +33,34 @@ func (r *ReconcileNodes) watchDeletions(stop <-chan struct{}) (chan string, erro
 	return chDels, nil
 }
 
+func (r *ReconcileNodes) watchUpdates() (chan string, error) {
```
lrgar, line 36: One option is to use the same informer as watchDeletions, and just return two channels, keeping a single connection. Don't have an opinion on that however.

pkg/controller/nodes/nodes_controller.go
```diff
@@ -288,3 +351,85 @@ func (r *ReconcileNodes) sendMarkedForTermination(oa *dynatracev1alpha1.OneAgent
 		},
 	})
 }
+
+func (r *ReconcileNodes) reconcileUnschedulableNode(node *corev1.Node, cache *Cache) error {
```
lrgar, line 355: Same here with `cache`

pkg/controller/nodes/nodes_controller.go
```diff
@@ -265,6 +311,23 @@ func (r *ReconcileNodes) removeNode(c *Cache, node string, oaFunc func(name stri
...
+	}
+
+	if isUnschedulable(node) {
+		err := r.reconcileUnschedulableNode(node, cache)
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
```
lrgar, line 328:
```suggestion
	if !isUnschedulable(node) {
		return nil
	}

	return r.reconcileUnschedulableNode(node, cache)
```

.gitignore
meik99: Yes, sometimes I start the operator with a redirection of its output into a log file. Since that log file does not belong into the repo i thought i add it to the gitignore so i don't accidentally commit it.

pkg/controller/nodes/oneagent_dao.go
meik99: data access object; my thought was that it connects the data from the cluster (e.g. data about OneAgent instances) with the code. First I wanted to name it "_controller", but we already have a file with that name. Yet, "_dao" is not that good of a name either, so I am open for suggestions.

PR #285, opened 2020-08-06
Update status of OneAgentAPM if token is missing

meik99:
Before returning the error, updates status of an OneAgentAPM instance if the token secret is missing.

### DIFF
--- a/pkg/controller/oneagentapm/oneagentapm_controller_test.go
+++ b/pkg/controller/oneagentapm/oneagentapm_controller_test.go
@@ -26,21 +26,26 @@ func init() {
 	os.Setenv(k8sutil.WatchNamespaceEnvVar, "dynatrace")
 }
 
-func TestReconcileOneAgentAPM(t *testing.T) {
-	name := "oneagent"
-	ns := "dynatrace"
+const (
+	apiUrl    = "https://ENVIRONMENTID.live.dynatrace.com/api"
+	name      = "oneagent"
+	namespace = "dynatrace"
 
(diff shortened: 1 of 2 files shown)


pkg/controller/oneagentapm/oneagentapm_controller_test.go
```diff
@@ -26,21 +26,26 @@ func init() {
...
 
-func TestReconcileOneAgentAPM(t *testing.T) {
-	name := "oneagent"
-	ns := "dynatrace"
+const (
+	apiUrl    = "https://ENVIRONMENTID.live.dynatrace.com/api"
+	name      = "oneagent"
+	namespace = "dynatrace"
+
+	noTokenError = "Secret 'dynatrace:oneagent' not found"
```
lrgar, line 34: If it's a variable used only once, then I suggest using it directly on the function who calls it.

pkg/controller/oneagentapm/oneagentapm_controller_test.go
```diff
@@ -26,21 +26,26 @@ func init() {
 	os.Setenv(k8sutil.WatchNamespaceEnvVar, "dynatrace")
 }
 
-func TestReconcileOneAgentAPM(t *testing.T) {
-	name := "oneagent"
-	ns := "dynatrace"
+const (
+	apiUrl    = "https://ENVIRONMENTID.live.dynatrace.com/api"
```
lrgar, line 30: `apiURL` please so that Go linter doesn't complain about it.

PR #287, opened 2020-08-06
removed log message

meik99:
Removed a log message which is not that important but spammed the log


pkg/controller/nodes/nodes_controller.go
```diff
@@ -101,9 +101,6 @@ func (r *ReconcileNodes) Start(stop <-chan struct{}) error {
 }
 
 func (r *ReconcileNodes) onUpdate(nodeName string) error {
-	logger := r.logger.WithValues("node", nodeName)
-	logger.Info("node update notification received")
```
lrgar, line 105: Why the removal? Does it log too much?

meik99: yes, on update seems to be called more often than anticipated, so the message spams the log, while not really saying anything of much use

PR #314, opened 2020-09-14
Feature/check min agent version

DTMad:


pkg/controller/utils/immutable_image.go
```diff
@@ -0,0 +1,42 @@
...
+		err := fmt.Errorf("dynatrace client is nil")
+		logger.Error(err, err.Error())
+		return false
+	}
+
+	if instance.GetSpec().UseImmutableImage &&
+		metav1.Now().UTC().Sub(instance.GetStatus().LastClusterVersionProbeTimestamp.UTC()) > updateInterval {
+		instance.GetStatus().LastClusterVersionProbeTimestamp = metav1.Now()
+		instance.GetStatus().UseImmutableImage =
+			instance.GetSpec().UseImmutableImage &&
```
lrgar, line 36: The upper if already checks for `instance.GetSpec().UseImmutableImage` to be true.

pkg/controller/utils/immutable_image.go
```diff
@@ -0,0 +1,42 @@
...
+	Reconcile(request reconcile.Request) (reconcile.Result, error)
+}
+
+// SetUseImmutableImageStatus sets the UseImmutableImage and LastClusterVersionProbeTimestamp stati of an BaseOneAgentDaemonSet instance
+// Returns true if:
+//     UseImmutableImage of specification is true &&
+//			LastClusterVersionProbeTimestamp status is the duration of updateInterval behind
+// otherwise returns false
+func SetUseImmutableImageStatus(logger logr.Logger, instance v1alpha1.BaseOneAgent, agentVersion string, dtc dtclient.Client) bool {
+	if dtc == nil {
```
lrgar, line 26: To be fair, we're ignoring whether the parameters can be nil pretty much everywhere.

pkg/controller/oneagentapm/oneagentapm_controller.go
```diff
@@ -43,6 +43,7 @@ func Add(mgr manager.Manager) error {
 		dtcReconciler: &utils.DynatraceClientReconciler{
 			Client:          client,
 			UpdatePaaSToken: true,
+			UpdateAPIToken:  true,
```
lrgar, line 46: Technically this is true only when immutable images are used. For customers upgrading they would receive errors for the missing ApiToken even if it doesn't affect them.

pkg/dtclient/cluster.go
```diff
@@ -0,0 +1,39 @@
...
+		_ = resp.Body.Close()
+	}()
+
+	responseData, err := dc.getServerResponseData(resp)
+	if err != nil {
+		return nil, err
+	}
+
+	err = json.Unmarshal(responseData, &result)
+	if err != nil {
```
lrgar, line 34:
```suggestion
	if err := json.Unmarshal(responseData, &result); err != nil {
```

pkg/version/agent_version_test.go
```diff
@@ -0,0 +1,57 @@
...
+	"github.com/stretchr/testify/assert"
+	"os"
+	logf "sigs.k8s.io/controller-runtime/pkg/runtime/log"
+	"testing"
+)
+
+func TestIsAgentVersionSupported(t *testing.T) {
+	logger := logf.ZapLoggerTo(os.Stdout, true)
+
+	t.Run("IsAgentVersionSupported", func(t *testing.T) {
```
lrgar, line 13: Why the subtest?

pkg/version/version_info.go
```diff
@@ -0,0 +1,68 @@
...
+		return result, nil
+	}
+
+	// Major and minor is equal, check release
+	result = a.release - b.release
+	return result, nil
+}
+
+func extractVersion(versionString string) (*versionInfo, error) {
+	mainVersionRegex := regexp.MustCompile(`[\d]+\.[\d]+\.[\d]+`)
```
lrgar, line 43: An example version is `1.203.12.20200914-092723` - I wonder if, for development pipelines, the first three fields remain the same, but the later changes.

Also, `regexp.MustCompile` is to be used as a global variable.

pkg/version/version_info.go
```diff
@@ -0,0 +1,68 @@
...
+	mainVersion := mainVersionRegex.FindString(versionString)
+
+	versionMergeRegex := regexp.MustCompile(`[\d]+`)
```
lrgar, line 46: Golang supports groups/submatches which is what you're trying to get here: https://golang.org/pkg/regexp/#Regexp.FindStringSubmatch

You could also just split the string by dot.

pkg/version/agent_version.go
```diff
@@ -0,0 +1,44 @@
...
+}
+
+func IsSupportedAgentVersion(agentVersion *versionInfo) (bool, error) {
+	if agentVersion == nil {
+		err := fmt.Errorf("parameter must not be nil")
+		return false, err
+	}
+
+	comparision, err := compareVersionInfo(agentVersion, minSupportedAgentVersion)
+	return comparision >= 0, err
```
lrgar, line 43:
```suggestion
	comparison, err := compareVersionInfo(agentVersion, minSupportedAgentVersion)
	return comparison >= 0, err
```

pkg/version/version_info.go
```diff
@@ -0,0 +1,68 @@
...
+	minor   int
+	release int
+}
+
+// compareVersionInfo returns:
+// 	0: if a == b
+//  n > 0: if a > b
+//  n < 0: if a < b
+//  0 with error: if a == nil || b == nil
+func compareVersionInfo(a *versionInfo, b *versionInfo) (int, error) {
```
lrgar, line 20: I'd suggest to remove the pointers, and just assume they are correct and leave the error verification problem to an upper level when it can actually be wrong.

meik99: pointers here are used to avoid pass-by-value due to higher memory consumption

pkg/version/version_info.go
meik99: Changed to FindStringSubmatch.

Splitting the string by dots returns 3 values for e.g.: 1.203.0 but 4 values for e.g.: 1.203.0.20200915-101926. Opted for regex to keep that more consistent.
Furthermore the used regex checks if the values between dots are actually numbers which makes checking for a malformed version string easier

pkg/controller/utils/immutable_image.go
```diff
@@ -0,0 +1,42 @@
...
+	"github.com/Dynatrace/dynatrace-oneagent-operator/pkg/version"
+	"github.com/go-logr/logr"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"time"
+)
+
+const updateInterval = 5 * time.Minute
+
+type Operator interface {
```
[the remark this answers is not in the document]
meik99: it is used to check if the last check has already happened in the past five minutes. Since the status is changed in the `SetUseImmutableImageStatus` function, the code triggers an instant reconcile and without this check it would end up in an  endless cycle of setting state and instantly reconciling

pkg/controller/istio/controller.go
```diff
@@ -325,7 +325,7 @@ func (c *Controller) createIstioConfigurationForVirtualService(oneagent dynatrac
 		return err
 	}
 	if vs == nil {
-		return fmt.Errorf("could not create virtual service with spec %v", virtualService.Spec)
+		return fmt.Errorf("Could not create virtual service with spec %v", virtualService.Spec)
```
lrgar, line 328:
```suggestion
		return fmt.Errorf("could not create virtual service with spec %v", virtualService.Spec)
```

go.sum
```diff
@@ -1,56 +1,91 @@
 bazil.org/fuse v0.0.0-20160811212531-371fbbdaa898/go.mod h1:Xbm+BRKSBEpa4q4hTSxohYNQpsxXPbPry4JJWOB3LB8=
```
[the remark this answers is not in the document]
lrgar: My point here remains.

pkg/controller/nodes/nodes_controller.go
```diff
@@ -14,7 +14,7 @@ import (
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
-	"k8s.io/apimachinery/pkg/runtime/schema"
+	schema "k8s.io/apimachinery/pkg/runtime/schema"
```
lrgar, line 17:
```suggestion
	"k8s.io/apimachinery/pkg/runtime/schema"
```

pkg/controller/nodes/nodes_controller_test.go
```diff
@@ -2,7 +2,7 @@ package nodes
 
 import (
 	"context"
-	"github.com/Dynatrace/dynatrace-oneagent-operator/pkg/apis"
+	apis "github.com/Dynatrace/dynatrace-oneagent-operator/pkg/apis"
```
lrgar, line 5:
```suggestion
	"github.com/Dynatrace/dynatrace-oneagent-operator/pkg/apis"
```

pkg/version/version_info.go
lrgar: These are supposed to be executed every 5 minutes, so that's not really a problem.

pkg/controller/oneagent/oneagent_controller_test.go
```diff
@@ -7,7 +7,7 @@ import (
 	"testing"
 	"time"
 
-	"github.com/Dynatrace/dynatrace-oneagent-operator/pkg/apis"
+	apis "github.com/Dynatrace/dynatrace-oneagent-operator/pkg/apis"
```
lrgar, line 10:
```suggestion
	"github.com/Dynatrace/dynatrace-oneagent-operator/pkg/apis"
```

pkg/controller/oneagentapm/oneagentapm_controller.go
```diff
@@ -97,8 +97,16 @@ func (r *ReconcileOneAgentAPM) Reconcile(request reconcile.Request) (reconcile.R
 		return reconcile.Result{}, errors.New(".spec.apiUrl is missing")
 	}
 
+	if instance.Spec.UseImmutableImage {
+		r.dtcReconciler.UpdateAPIToken = true
+	}
+
 	dtc, upd, err := r.dtcReconciler.Reconcile(context.TODO(), instance)
```
lrgar, line 104: Technically `r.dtcReconciler` is at risk of race conditions if multiple OneAgentAPM `Reconcile` calls are running in parallel. I think by default running reconciliation in parallel is disabled so it's not a problem now, but I'd suggest the changes below (in part to avoid changes on global objects),

```suggestion
	dtcRec := *r.dtcReconciler
	if instance.Spec.UseImmutableImage {
		dtcRec.UpdateAPIToken = true
	}

	dtc, upd, err := dtcRec.Reconcile(context.TODO(), instance)
```

pkg/controller/oneagentapm/oneagentapm_controller_test.go
```diff
@@ -5,7 +5,7 @@ import (
 	"os"
 	"testing"
 
-	"github.com/Dynatrace/dynatrace-oneagent-operator/pkg/apis"
+	apis "github.com/Dynatrace/dynatrace-oneagent-operator/pkg/apis"
```
lrgar, line 8:
```suggestion
	"github.com/Dynatrace/dynatrace-oneagent-operator/pkg/apis"
```

pkg/dtclient/communication_hosts_test.go
```diff
@@ -166,7 +166,7 @@ func handleCommunicationHosts(request *http.Request, writer http.ResponseWriter)
 	switch request.Method {
 	case "GET":
 		writer.WriteHeader(http.StatusOK)
-		_, _ = writer.Write(commHostOutput)
+		_, _ = writer.Write([]byte(commHostOutput))
```
lrgar, line 169:
```suggestion
		_, _ = writer.Write(commHostOutput)
```

pkg/webhook/bootstrapper/controller_test.go
```diff
@@ -19,6 +19,8 @@ import (
 	logf "sigs.k8s.io/controller-runtime/pkg/runtime/log"
 )
 
+const testNamespace = "dynatrace"
+
```
lrgar, line 23: Unused.

pkg/version/version_info.go
```diff
@@ -0,0 +1,67 @@
...
+	// Major and minor is equal, check release
+	result = a.release - b.release
+	return result, nil
+}
+
+func extractVersion(versionString string) (*versionInfo, error) {
+	version := versionRegex.FindStringSubmatch(versionString)
+
+	if len(version) < 4 {
+		return nil, fmt.Errorf("version malformed: %s", versionString)
```
lrgar, line 48: Just in case, things such as `versionRegex.FindStringSubmatch("asdadasdsd.asd1.2.3")` would pass the validation, although they most certainly won't happen.

pkg/controller/utils/immutable_image.go
```diff
@@ -0,0 +1,41 @@
...
+		return false
+	}
+
+	// Variable declared to make if-condition more readable
+	lastClusterVersionProbeTimestamp := instance.GetStatus().LastClusterVersionProbeTimestamp.UTC()
+	if instance.GetSpec().UseImmutableImage && isLastProbeOutdated(lastClusterVersionProbeTimestamp) {
+		instance.GetStatus().LastClusterVersionProbeTimestamp = metav1.Now()
+		instance.GetStatus().UseImmutableImage =
+			version.IsRemoteClusterVersionSupported(logger, dtc) &&
+				version.IsAgentVersionSupported(logger, agentVersion)
```
lrgar, line 33: There is also the OneAgent version configured on the cluster.

For example, customer installing Operator in a new environment using our instructions and we have `.spec.useImmutableImage` enabled by default in our example, their cluster supports images, but the OneAgent version they want to use (and **it's configured on Dynatrace**, i.e., not on .spec) doesn't, we need to disable immutable image for this case.

In fact the version on `.spec` is not particularly relevant since we don't modify the installer URL anyway if the version is unsupported.

This may not be an uncommon situation at the beginning once we release this feature.

pkg/controller/utils/immutable_image.go
```diff
@@ -0,0 +1,46 @@
...
+	}
+
+	// Variable declared to make if-condition more readable
+	lastClusterVersionProbeTimestamp := instance.GetStatus().LastClusterVersionProbeTimestamp.UTC()
+	if instance.GetSpec().UseImmutableImage && isLastProbeOutdated(lastClusterVersionProbeTimestamp) {
+		instance.GetStatus().LastClusterVersionProbeTimestamp = metav1.Now()
+		agentVersion, err := dtc.GetLatestAgentVersion(dtclient.OsUnix, dtclient.InstallerTypeDefault)
+		if err == nil {
+			instance.GetStatus().UseImmutableImage =
+				version.IsRemoteClusterVersionSupported(logger, dtc) &&
```
lrgar, line 34: I presume the issue with agent version failing also applies to fetching the remote cluster version, doesn't it? That is, a network glitch may disable immutable images temporarily.

lrgar (approved):

PR #331, opened 2020-10-12
Remove OneAgentIM and add webhookInjection property

DTMad:


deploy/cr.yaml
```diff
@@ -41,6 +41,11 @@ spec:
   #
   # useImmutableImage: true
 
+  # [On development, to be released]
+  # If enabled this will automatically use the webhook for injecting into applications
```
lrgar, line 45: No, it won't. It will only stop the OneAgent Pod from doing it. The customer still needs to use OneAgentAPM.

Also, do we want this here now? It's still a work in progress on the agent/cluster side.

deploy/crds/dynatrace.com_oneagents_crd.yaml
```diff
@@ -221,6 +221,7 @@ spec:
                   type: object
                 requests:
                   additionalProperties:
+                    anyOf:
```
lrgar, line 224: Weird line.

pkg/apis/dynatrace/v1alpha1/oneagent_types.go
```diff
@@ -60,6 +51,12 @@ type OneAgentSpec struct {
 	// +operator-sdk:gen-csv:customresourcedefinitions.specDescriptors.x-descriptors="urn:alm:descriptor:com.tectonic.ui:text"
 	CustomPullSecret string `json:"customPullSecret,omitempty"`
 
+	// Enables automatic injection into applications with the webhook
```
lrgar, line 54: See my comment on cr.yaml

pkg/controller/oneagent/oneagent_controller.go
```diff
@@ -50,12 +50,12 @@ func Add(mgr manager.Manager) error {
 		mgr.GetConfig(),
 		log.Log.WithName("oneagent.controller"),
 		utils.BuildDynatraceClient,
-		&dynatracev1alpha1.OneAgent{}))
+		dynatracev1alpha1.OneAgent{}))
```
lrgar, line 53: This argument is no longer needed.

pkg/controller/oneagent/oneagent_controller_test.go
```diff
@@ -79,7 +79,7 @@ func TestReconcileOneAgent_ReconcileOnEmptyEnvironmentAndDNSPolicy(t *testing.T)
 			UpdatePaaSToken:     true,
 			UpdateAPIToken:      true,
 		},
-		instance: &dynatracev1alpha1.OneAgent{},
+		instance: dynatracev1alpha1.OneAgent{},
```
lrgar, line 82: This field shouldn't be needed anymore.

lrgar (changes requested):

deploy/crds/dynatrace.com_oneagents_crd.yaml
DTMad: Removed

pkg/controller/oneagent/oneagent_controller_test.go
DTMad: Removed

pkg/controller/oneagent/oneagent_controller.go
DTMad: Removed

lrgar (approved):

PR #332, opened 2020-10-13
Add configurable resource setting for OneAgentAPM

DTMad:


deploy/cr-apm.yaml
```diff
@@ -51,3 +51,14 @@ spec:
   # Sets a NetworkZone for the OneAgent (optional)
   # Note: This feature requires OneAgent version 1.195 or higher
   #networkZone: name-of-my-network-zone
+
+  # Optional: resource settings for initContainer. Consumption of the OneAgent heavily depends
```
lrgar, line 55: Not really, this the consumption is pretty much static. This is only for the unzipping.

Well, I guess if your OneAgent package is bigger, the unzipping can need more CPU.

pkg/apis/dynatrace/v1alpha1/base_types.go
```diff
@@ -78,6 +79,12 @@ type BaseOneAgentSpec struct {
 	// +operator-sdk:gen-csv:customresourcedefinitions.specDescriptors.displayName="Use immutable image"
 	// +operator-sdk:gen-csv:customresourcedefinitions.specDescriptors.x-descriptors="urn:alm:descriptor:com.tectonic.ui:advanced,urn:alm:descriptor:com.tectonic.ui:booleanSwitch"
 	UseImmutableImage bool `json:"useImmutableImage,omitempty"`
+
+	// Optional: define resources requests and limits for single pods
```
lrgar, line 83:
```suggestion
	// Optional: define resources requests and limits for the attached init container
```

You may need to have the field individually for OneAgent and OneAgentAPM (and to be fair, I wonder if we should rename it for the OneAgentAPM since they don't mean the same)

pkg/webhook/server/server.go
```diff
@@ -175,6 +174,11 @@ func (m *podInjector) Handle(ctx context.Context, req admission.Request) admissi
 		useImmutableImage = "true"
 	}
 
+	resources := oa.Spec.Resources
+	if resources.Requests == nil {
+		resources.Requests = corev1.ResourceList{}
```
lrgar, line 179: Why this line?

lrgar (approved):

PR #333, opened 2020-10-13
Configure unprivileged mode on OneAgent CR

lrgar:
The feature was previously configured through an environment variable. It's now the `useUnprivilegedMode` field on the OneAgent Custom Resource.


pkg/apis/dynatrace/v1alpha1/oneagent_types.go
```diff
@@ -109,6 +109,12 @@ type OneAgentSpec struct {
 	// +operator-sdk:gen-csv:customresourcedefinitions.specDescriptors.displayName="Labels"
 	// +operator-sdk:gen-csv:customresourcedefinitions.specDescriptors.x-descriptors="urn:alm:descriptor:com.tectonic.ui:advanced,urn:alm:descriptor:com.tectonic.ui:text"
 	Labels map[string]string `json:"labels,omitempty"`
+
+	// Optional: Runs the OneAgent Pods as unprivileged (Early Preview)
```
DTMad, line 113: I think the official term we use is 'Early Adopter'

lrgar: You're right. Will fix.

DTMad (approved):

PR #335, opened 2020-10-19
Add disable container injection env var

DTMad:


lrgar (approved):
Squash when merging.

Plus only merge once 0.9.0 has been released.

meik99 (approved):

PR #340, opened 2020-11-09
set node properties conditionally

meik99:


pkg/controller/namespace/namespace_controller.go
```diff
@@ -301,11 +301,16 @@ k8s_poduid ${K8S_PODUID}
 k8s_containername ${container_name}
 k8s_basepodname ${K8S_BASEPODNAME}
 k8s_namespace ${K8S_NAMESPACE}
+EOF
 {{- if .AddNodeProps}}
+	if [[ "${api_url}" == *"${host_tenant}"* ]] && [[ "${im_nodes[${K8S_NODE_NAME}]+0}" ]]; then
```
lrgar, line 306: Don't use API URL here, `{{.OneAgent.Status.EnvironmentID}}` will give you the tenant. You don't need the second condition either since `host_tenant` wouldn't be set if it's not a host node. Hence this can be added to the `if [[ ! -z "${host_tenant}" ]]; then` below which is already checks half of the conditions.

lrgar (changes requested):

pkg/controller/namespace/namespace_controller.go
```diff
@@ -301,13 +301,17 @@ k8s_poduid ${K8S_PODUID}
 k8s_containername ${container_name}
 k8s_basepodname ${K8S_BASEPODNAME}
 k8s_namespace ${K8S_NAMESPACE}
+EOF
+
 {{- if .AddNodeProps}}
+	if [[ ! -z "${host_tenant}" ]]; then		
+		if [[ "{{.OneAgent.Status.EnvironmentID}}" == "${host_tenant}" ]]; then
+					cat <<EOF >>${container_conf_file}
```
lrgar, line 309: A bit of extra indentation, but good otherwise.

lrgar (approved):

PR #341, opened 2020-11-10
add libc property

meik99:


deploy/cr-apm.yaml
```diff
@@ -49,3 +50,8 @@ spec:
   #   limits:
   #     cpu: 500m
   #     memory: 100M
+
+  # Optional: define which core library for C should be used.
```
baichinger, line 54: Could be rephrased to "defines the C standard library used."

pkg/apis/dynatrace/v1alpha1/oneagentapm_types.go
```diff
@@ -45,6 +45,13 @@ type OneAgentAPMSpec struct {
 	// +operator-sdk:gen-csv:customresourcedefinitions.specDescriptors.displayName="Resource Requirements"
 	// +operator-sdk:gen-csv:customresourcedefinitions.specDescriptors.x-descriptors="urn:alm:descriptor:com.tectonic.ui:advanced,urn:alm:descriptor:com.tectonic.ui:resourceRequirements"
 	Resources corev1.ResourceRequirements `json:"resources,omitempty"`
+
+	// Optional: define which core library for C should be used.
+	// Can be set to "musl" to use musl instead of glibc
+	// +operator-sdk:gen-csv:customresourcedefinitions.specDescriptors=true
+	// +operator-sdk:gen-csv:customresourcedefinitions.specDescriptors.displayName="C Core Library"
```
baichinger, line 52: Please use _C standard library_ as well.

pkg/webhook/server/flavor.go
```diff
@@ -0,0 +1,16 @@
...
+	"github.com/Dynatrace/dynatrace-oneagent-operator/pkg/controller/utils"
+	dtwebhook "github.com/Dynatrace/dynatrace-oneagent-operator/pkg/webhook"
+)
+
+func getFlavor(libc string /*oa.Spec.LibC*/, annotations map[string]string /* pod.Annotations */) string {
+	flavor := url.QueryEscape(utils.GetField(annotations, dtwebhook.AnnotationFlavor, libc))
+	if flavor != "musl" {
+		return "default"
+	}
+	return flavor
```
baichinger, line 15:
```suggestion
	if flavor = "musl" {
		return flavor
	}
	return "default
```

baichinger (commented):
Will this cause confusion when the user needs to know that annotation _flavor_ is actually the same as custom resource field _libc_. It will get even more strange when values of them do not match: `musl` vs. `default` vs. `glibc`.

Therefore, shouldn't we simply stick with the term _flavor_?

lrgar (approved):

DTMad (approved):

PR #358, opened 2021-01-08
Change image pull policy for OneAgentAPM with useImmutableImage

DTMad:


pkg/webhook/server/server.go
```diff
@@ -170,15 +170,18 @@ func (m *podInjector) Handle(ctx context.Context, req admission.Request) admissi
 	}
 
 	useImmutableImage := ""
+	imagePullPolicy := corev1.PullIfNotPresent
 	if oa.Status.UseImmutableImage {
 		useImmutableImage = "true"
+		imagePullPolicy = corev1.PullAlways
 	}
```
lrgar, line 177: Always set it to PullAlways, there is no reason not to. If you don't use the immutable image, then we use the Operator image, and we want to use the same one used by the Operator Pod.

lrgar (changes requested):

pkg/webhook/server/server.go
DTMad: Done

lrgar (approved):

PR #360, opened 2021-01-14
Cache Docker image queries

lrgar:


pkg/controller/oneagent/oneagent_update.go
```diff
@@ -174,14 +175,39 @@ func (r *ReconcileOneAgent) findOutdatedPodsImmutableImage(logger logr.Logger, i
...
+
+		dockerVersionChecker := utils.NewDockerVersionChecker(image, imageID, dockerConfig)
+		latest, err := dockerVersionChecker.IsLatest()
+		if err != nil {
+			failed = true
+			logger.Info(err.Error())
+			return true, nil
+		}
+
+		cache[imageID] = latest
```
DTMad, line 208: is there a case where imageID can be nil?

DTMad (approved):

pkg/controller/oneagent/oneagent_update.go
lrgar: It's checked beforehand: https://github.com/Dynatrace/dynatrace-oneagent-operator/blob/78481d65deb6d69843f3ca00825a6c2fa58b2aaf/pkg/controller/oneagent/oneagent_update.go#L144-L161

PR #381, opened 2021-03-19
default pull secret name changed

meik99:
appended clusterrole for operator due to default pull secret name change


DTMad (approved):

lrgar:
Why did it change? If there was no reason, then it's a bug.

PR #388, opened 2021-07-23
Change pull policy to ifnotpresent

rspierz:
Allow image to be pulled directly from node instead of pulled from registry each time. Save on pull time and also provides a level of workflow if registry is unavailable.


rspierz:
@DTMad Could you have a look at this PR?

